PR #12, opened 2017-02-19
Add AsType codec from Zarr

jakirkham:
Fixes https://github.com/alimanfoo/zarr/issues/102
Fixes https://github.com/alimanfoo/numcodecs/issues/7
Incorporates https://github.com/alimanfoo/zarr/pull/120

~~This is very much WIP. In particular, the tests need a bit of help getting cleaned up.~~ Resolved.

This adds the AsType codec from Zarr to numcodecs.

### DIFF
--- a/numcodecs/__init__.py
+++ b/numcodecs/__init__.py
@@ -49,6 +49,9 @@ try:
 except ImportError:  # pragma: no cover
     pass
 
+from numcodecs.astype import AsType
+register_codec(AsType)
+
 from numcodecs.delta import Delta
 register_codec(Delta)
 
--- a/numcodecs/astype.py
+++ b/numcodecs/astype.py
@@ -1,0 +1,88 @@
+# -*- coding: utf-8 -*-
+from __future__ import absolute_import, print_function, division
+
+import numpy as np
+
+from numcodecs.abc import Codec
+from numcodecs.compat import buffer_copy, ndarray_from_buffer
+
+
+
+class AsType(Codec):
+    """Filter to convert data between different types.
+
+    Parameters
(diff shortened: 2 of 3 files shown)


alimanfoo:
Thanks @jakirkham. The `astype` module looks fine, just needs some work on the tests. There is a `numcodecs.tests.common` module which is supposed to provide some general support for encoding and decoding checks for any codec, it would be good to use as in the other tests modules. If you have time to spend that would be great, but just say if not, I'd be happy to merge as-is and then work on afterwards.

jakirkham:
Thanks for the tips.

Went ahead and borrowed from the test suite for `Delta` in Numcodecs (much as was done in Zarr). Not sure if I have covered all the cases that were covered in Zarr, but don't have much more time to explore this near term. In any event, the tests seem to be passing and expanding the test suite should be straightforward now that there are some tests in place. 😉

Also have incorporated some changes to the representation of `AsType` that are the same as one's I'm proposing to Zarr. Hope that is ok. They were modeled after similar changes to `Delta`, which I'm guessing came about to handle Endianness. As such, it seems like an important change to include in both. Hope it is ok that I rolled it into here.

alimanfoo:
That's great, thank you. I'm  happy to merge this and review the tests afterwards.

jakirkham:
SGTM

alimanfoo:
Thanks very much.

jakirkham:
Of course. Thought it was the least I could do after adding it to Zarr while in the midst of this transition.

Will try to go back and look at the tests in Zarr as well when I have more time.

PR #58, opened 2017-12-06
Fix VLenBytes docstring to use VLenBytes

jakirkham:
Had a copy-paste error where it was using `VLenUTF8` previously.


jakirkham:
Just manually edited the C file. Roughly checks out when I created it with Cython, but was using a different version and probably not the same settings used to create the C file originally.

alimanfoo:
Thanks, I've manually merged in #57. Also I've added running of doctests for the cython modules to tox so hopefully should catch this kind of thing automatically in future.

jakirkham:
Ok, thanks.

PR #71, opened 2018-03-05
Upgrade c-blosc to v1.14.0

alimanfoo:
This PR upgrades the bundled c-blosc sources to v1.14.0.

* [x] ``tox -e py36`` passes locally

### DIFF
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -1,6 +1,15 @@
 Release notes
 =============
 
+
+.. _release_0.5.4:
+
+0.5.4
+-----
+
+* The bundled c-blosc sources have been upgraded to version 1.14.0 (:issue:`71`).
+
+
 .. _release_0.5.3:
 
 0.5.3
@@ -12,6 +21,7 @@ Release notes
 * The bundled c-blosc library has been updated to version 1.13.4 (:issue:`63`,
   :issue:`64`).
 
+
 .. _release_0.5.2:
 
 0.5.2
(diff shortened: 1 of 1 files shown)


jakirkham:
FWIW needed a patch to build this on VS 2008, which is in PR ( https://github.com/Blosc/c-blosc/pull/221 ). Not sure if we will need it here. However if we do, can just move up to that one commit beyond the tag to grab it.

jakirkham:
Looks like the patch wasn't needed after all. Great. Guessing that only affected building/running tests.

Anything else needed here? Just looks like the changelog entry?

alimanfoo:
Just the changelog, have pushed that. Should be good to go.

jakirkham:
Sorry @alimanfoo. It's been a hectic day. This seems fine to me.

That said, it doesn't seem to allow me to merge as coveralls doesn't show up. Am almost certain this is a communication issue between the CIs and Coveralls. Could give one of the CI builds a restart to see if that fixes the issue. Thoughts?

alimanfoo:
Thanks @jakirkham, I think I'll just override and merge, the latest commit is docs only and I meant to ``[ci skip]`` but forgot, previous commit had 100% coverage.

alimanfoo:
Also you should have admin access to the repo now.

jakirkham:
SGTM. Thanks @alimanfoo.

GitHub use to not be very smart about `[ci skip]` and required CIs. Namely it would still block a PR unless those CIs showed up (even though that would clearly not happen with `[ci skip]`). Not sure if they fixed this issue since.

Ah ok. Thanks.

alimanfoo:
Yeah I think github is still like that, if I had done ``[ci skip]`` it would still have blocked the PR.

PR #82, opened 2018-05-23
Adding support for ppc64le architecture

anandtrex:
This is done just by updating the cpuinfo.py file from upstream

TODO:
* [x] ``tox -e py36`` passes locally

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -29,7 +29,9 @@ have_sse2 = 'sse2' in flags
 have_avx2 = 'avx2' in flags
 disable_sse2 = 'DISABLE_NUMCODECS_SSE2' in os.environ
 disable_avx2 = 'DISABLE_NUMCODECS_AVX2' in os.environ
-
+if PY2 and os.name == 'nt':
+    # force no AVX2 on windows PY27
+    disable_avx2 = True
 
 # setup common compile arguments
 have_cflags = 'CFLAGS' in os.environ
@@ -105,7 +107,7 @@ def blosc_extension():
         info('compiling Blosc extension with AVX2 support')
         extra_compile_args.append('-DSHUFFLE_AVX2_ENABLED')
         blosc_sources += [f for f in glob('c-blosc/blosc/*.c') if 'avx2' in f]
-        if os.name == 'nt' and not PY2:
+        if os.name == 'nt':
             define_macros += [('__AVX2__', 1)]
     else:
         info('compiling Blosc extension without AVX2 support')


alimanfoo:
Thank you @anandtrex for this. It looks like some of the Windows builds are failing to compile blosc, I haven't seen that before. I've kicked off a rebuild, will see what happens.

anandtrex:
@alimanfoo Any ideas on why the windows build fails? I have no experience with windows at all, so I'm having some trouble interpreting the error messages.

anandtrex:
Hi @alimanfoo Yes, in my PR, the cpuinfo.py is identical to the upstream one.

alimanfoo:
Thinking about this some more, we should be able to install cpuinfo from pypi by including it in setup_requires. This would avoid having to vendor cpuinfo ourselves. I think it would be worth giving it a try, should reduce the maintenance burden in future.

alimanfoo:
Regarding the CI failures, it might be worth waiting until #92 is merged then rebasing and seeing if any build failures remain.

alimanfoo:
@zarr-developers/core-devs any objections to adding ``cpuinfo`` into ``setup_requires`` rather than vendoring?

jakirkham:
IME `setup_requires` can be a bit flaky. Namely `pip` does not really support it and some `setuptools` versions have broken it. That said, I have used it before and if that seems like the best solution, we can do it.

A better solution would be to follow [PEP 518]( https://www.python.org/dev/peps/pep-0518/ ).

alimanfoo:
Yeah, setup_requires is such a mess. We're using it currently for setuptools-scm, so I think I figured it kinda works, but I could be wrong. Is PEP 518 implemented yet?

alimanfoo:
> Yeah, setup_requires is such a mess. We're using it currently for setuptools-scm, so I think I figured it kinda works, but I could be wrong. Is PEP 518 implemented yet?

Answering my own question, looks like it is [fully supported](https://pip.pypa.io/en/stable/news/) as of pip 10.

alimanfoo:
Actually I wonder if PEP518 is implemented maybe we should give it a try.

jakirkham:
It probably is worth trying. Was just trying to avoid feature creep within this PR. 😄

alimanfoo:
Hi all, I've made a PR (#130) which uses setup_requires for cpuinfo. I tried using pyproject.toml but there are complications and I think it's worth waiting for further work to be completed upstream (e.g., in pip and tox). My proposal is to merge #130 instead of this PR for ease of maintenance. Please let me know if any objections.

alimanfoo:
I've backed off #130 and think for the current release would be better to pursue this PR. I've merged in latest changes from master, will see if CI failures recur.

alimanfoo:
OK, travis is green, but appveyor is failing on PY27. Problem is with blosc compilation, this is the error:

```
c:\projects\numcodecs\c-blosc\blosc\blosc-common.h(77) : fatal error C1083: Cannot open include file: 'immintrin.h': No such file or directory
```

Looks like this issue is similar: https://github.com/facebook/zstd/issues/9

jakirkham:
If we make sure `__AVX2__` is undefined on Windows Visual Studio 2008, that should work.

ref: https://github.com/Blosc/c-blosc/search?q=immintrin.h&unscoped_q=immintrin.h

alimanfoo:
Thanks @jakirkham. So would that be the same as, in setup.py, replacing:

```python
        if os.name == 'nt':
            define_macros += [('__AVX2__', 1)]
```

...with:

```python
        if os.name == 'nt' and not PY2:
            define_macros += [('__AVX2__', 1)]
```

...?

jakirkham:
Sure, that seems reasonable.

alimanfoo:
OK, cool, I've disabled AVX2 on PY2 windows, if CI passes I'll merge.

alimanfoo:
Trying a better way to disable AVX2 on windows PY2...

PR #87, opened 2018-10-18
Add GZip codec

funkey:
This adds a dedicated gzip codec, `GZip` with ID `gzip`. This replaces the previous alias `zlib` for `gzip`, which did not take into account that the compression headers between zlib and gzip differ. In particular, `ZLib` could not be used to decode data compressed using gzip by a third party (which would contain a gzip header).

For backwards compatibility with archives that have been compressed with `gzip` (and therefore `ZLib`), `GZip` supports both header variants for decoding.

TODO:


numcodecs/gzip.py
```diff
@@ -0,0 +1,68 @@
...
+            # if numpy array, can only handle C contiguous directly
+            if not buf.flags.c_contiguous:
+                buf = buf.tobytes(order='A')
+
+        if PY2:  # pragma: py3 no cover
+            # ensure bytes, PY2 cannot handle things like bytearray
+            buf = buffer_tobytes(buf)
+
+        # do compression
+        compress = _zlib.compressobj(self.level, wbits=16 + 15)
```
jakirkham: Looks like Python 2 doesn't support keyword arguments. Fortunately it does support `wbits`, it just comes as the third positional argument. Something like `_zlib.compressobj(self.level, _zlib.DEFLATED, 16 + 15)` should do the trick and work on both Python 2 and Python 3.

jakirkham:
Thanks for working on this, Jan. Needs a few tweaks for CI, but otherwise seems fine to me. Let's see if @alimanfoo has anything to add.

Note: There appears to be a `test_alias` in `numcodecs/tests/test_zlib.py`, which needs to be updated (or removed?) so as to not treat `gzip` as an alias.

funkey:
Yes, help with the CI would be very appreciated. I couldn't get `tox` to run locally, there are probably a few locations that need some attention.

jakirkham:
Added PR ( https://github.com/funkey/numcodecs/pull/1 ) against this PR, which should fix the test issues.

jakirkham:
Do we need a fixture for this as well? Is there anything special we need to do to generate it?

alimanfoo:
The first run of [test_backwards_compatibility()](https://github.com/zarr-developers/numcodecs/pull/87/files#diff-b900e8c5a086c4b9a2ed4357e78a51f8R73) should generate a fixture. That fixture (files inside the fixture directory) should be included in this PR. Subsequent runs of that test will then compare with the fixture files to ensure nothing has changed.

jakirkham:
Does one need specific version of dependencies in their environment to do this correctly?

numcodecs/tests/test_gzip.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+def test_repr():
+    check_repr("GZip(level=3)")
+
+
+def test_alias():
+    config = dict(id='gzip', level=1)
+    codec = get_codec(config)
+    assert GZip(1) == codec
```
jakirkham: Guess we should drop this one as well.

xref: https://github.com/funkey/numcodecs/pull/1#pullrequestreview-166439611

jakirkham:
Submitted PR ( https://github.com/funkey/numcodecs/pull/2 ) against this PR, which drops the alias tests, commits the `gzip` fixture (used Python 3.6), and fixes the flake8 error that CI is reporting.

jakirkham:
Looks like CI is now green. 😄

The API docs can probably just copy `docs/zlib.rst` and rename some things. The release notes are in `docs/release.rst`, which could use a quick sentence on this (unless you want to write more).

alimanfoo:
Just wondering, do we want to be pushing users towards gzip rather than
zlib? Is one more broadly compatible than the other?

numcodecs/tests/test_gzip.py
```diff
@@ -0,0 +1,75 @@
...
+    np.random.choice([b'a', b'bb', b'ccc'], size=1000).reshape(10, 10, 10),
+    np.random.randint(0, 2**60, size=1000, dtype='u8').view('M8[ns]'),
+    np.random.randint(0, 2**60, size=1000, dtype='u8').view('m8[ns]'),
+    np.random.randint(0, 2**25, size=1000, dtype='u8').view('M8[m]'),
+    np.random.randint(0, 2**25, size=1000, dtype='u8').view('m8[m]'),
+]
+
+
+def test_encode_decode():
+    for arr, codec in itertools.product(arrays, codecs):
```
jakirkham: No need to change. Just noting for reference. This can be done with [`pytest.mark.parametrize`]( https://docs.pytest.org/en/latest/parametrize.html#pytest-mark-parametrize-parametrizing-test-functions ).

jakirkham:
Docs look good. Thanks for adding those.

jakirkham:
Started look at this again trying to figure out an answer to @alimanfoo's question.

Noticed that GZip has a footer (with some checksum info) and optional headers (assuming I'm reading [this spec]( http://www.onicos.com/staff/iz/formats/gzip.html ) right and found the right spec). This could be valuable information to have and would justify using GZip over Zlib. That said, we don't currently handle this info it seems and it might be a bit tedious to do manually.

Looking at Python 3 it seems they include `compress` and `decompress` functions ( https://github.com/python/cpython/commit/79c5ef11d56e761445e334a7121764808e9e70cf ) by just using a `GzipFile` instance with a `fileobj` based off a `BytesIO` instance. While Python 2 doesn't have these functions, it looks like we could implement these ourselves with little difficulty in a Python 2/3 compatible way. This should handle these other header and footer fields and maybe even validate the checksum for us.

Thoughts?

alimanfoo:
Re using gzip.compress (i.e., GzipFile) versus zlib.compressobj, happy to defer judgement. Couple of thoughts for consideration...

Checking the CRC might be a nice feature. Are there any performance costs? I.e., would users always/sometimes/never want to check the CRC?

Are there any other performance differences between using GzipFile versus zlib.compressobj directly, for either compression or decompression?

jakirkham:
Just to clarify the Python `GzipFile` object always uses CRC32 from Zlib. AFAICT there is no option to disable this. My reading of the Gzip spec (and please feel free to correct me if I'm wrong) is that including the CRC32 is a required part of the footer. Not sure how tools handle cases where this is not provided or the value is incorrect, but guessing it varies from erroring out (ideally nicely) to gracefully handling this somehow (maybe a warning). Python specifically raises an `OSError`.

There are some other things that the Gzip spec requires like a remainder of the decompressed file size in the footer. Some optional fields are available as well (file name, file comment, etc.), which we probably don't care about using ourselves, but it would be good to handle them correctly (in the event we or others opt to use them for some reason).

So my take on this is the choice is not between whether to include CRC32 in the Gzip encoding, but whether we let Python do the heavy lifting for us or whether we do it manually ( personally would prefer the former :) ).

jakirkham:
> Checking the CRC might be a nice feature. Are there any performance costs? I.e., would users always/sometimes/never want to check the CRC?

So admittedly I'm not an expert on this, but after a bit of reading here's what I have come up with. CRCs are the go to error correction code. While there are others (e.g. Fletcher and Adler), their ability to catch errors pales in comparison to CRCs. The CRC algorithm is based on polynomial division, which makes it effective at catching errors, but also makes it slow compared to the alternatives.

Research has been done to find a few good polynomials performance-wise that pretty much all CRC algorithms use one of. There are various implementations (many using precomputed tables as Zlib appears to) and some chips (Intel with SSE4.2 support) have intrinsics that utilize a hardware implementation of the algorithm. Performance varies depending on the size of the data, whether it is a hardware or software implementation, whether a table is used, how large a chunk of data is processed, which polynomial is used, and compiler optimizations.

Here's a benchmark of [common CRC implementations]( https://fastmail.blog/2015/12/03/the-search-for-a-faster-crc32/ ). It's worth noting that mainly networking types are interested in high performance here as they are running CRC all the time. It's unclear to me whether or not it would be that much of a concern for our use cases. However this could be a deciding factor between Gzip and Zlib encodings for our users (or perhaps motivation to pick up a Zlib patched for better performance like [Intel's]( https://www.intel.com/content/dam/www/public/us/en/documents/white-papers/zlib-compression-whitepaper-copy.pdf ) or [zlib-ng]( https://github.com/Dead2/zlib-ng ) if using CRC is important).

alimanfoo:
Sorry for lack of clarity, I was meaning that when we decompress we have a
choice about whether to check the CRC. IIUC the current implementation in
this PR does *not* check the CRC. If we switched to using gzip.decompress()
(which uses GzipFile.read()) then the CRC would *always* get checked.

Computing the CRC to check against the footer when decompressing could have
some noticeable performance overhead, or might be negligible, I don't know.
If there was some overhead, a user might want to turn it off, i.e., don't
bother to compute the CRC and check. If not, and performance was equal
between different possible implementations under discussion, then we might
prefer to use the gzip module because it includes a CRC check and that
doesn't hurt.

I'm happy to defer to you on this, just trying to identify things to
consider.

alimanfoo:
SGTM

funkey:
@jakirkham: Thanks a lot for your investigation. Using CRC checks during reading as a default sounds good to me, too.

jakirkham:
Thanks all.

Seems like CIs are mostly happy with the exception of Python 3.4. Think we are hitting some edge case of that older Python version that the other Pythons either fixed or didn't have in the first place. Did a little bit of digging, but wasn't able to find the exact issue that was fixed in Python. Was debating remedying it; however, we have dropped Python 3.4 from Zarr just recently. So wondering if we should just do the same here. Proposing just dropping Python 3.4 in PR ( https://github.com/zarr-developers/numcodecs/pull/89 ).

alimanfoo:
Just checking, was the fixture data regenerated after switching to the GzipFile implementation? Looks like it was but on a mobile and wanted to double check.

jakirkham:
Yep, they were regenerated.

Had to explicitly delete the existing gzip fixtures (added earlier in this PR) and rerun the tests to regenerate them. Not sure if that is expected. In any event, seemed to work otherwise.

jakirkham:
Also just ran one chunk of encoded data through the `gzip` CLI. First wanted to verify that `gzip` could use it without issues. Second wanted to manually verify the NumPy data used to generate it matched the binary decompressed file's contents post-`gzip`. Both of these worked correctly without issues.

alimanfoo:
>
> Yep, they were regenerated.
>
> Had to explicitly delete the existing gzip fixtures (added earlier in this
> PR) and rerun the tests to regenerate them. Not sure if that is expected.
> In any event, seemed to work otherwise.
>
Great, yes that is the process.

-- 
Please feel free to resend your email and/or contact me by other means if
you need an urgent reply.

Alistair Miles
Head of Epidemiological Informatics
Centre for Genomics and Global Health
Big Data Institute
Li Ka Shing Centre for Health Information and Discovery
Old Road Campus
Headington
Oxford
OX3 7LF
United Kingdom
Phone: +44 (0)1865 743596 or +44 (0)7866 541624
Email: alimanfoo@googlemail.com
Web: http://a <http://purl.org/net/aliman>limanfoo.github.io/
Twitter: @alimanfoo <https://twitter.com/alimanfoo>

jakirkham:
Any other thoughts on this or are we ready to merge?

docs/release.rst
```diff
@@ -20,6 +20,8 @@ Release notes
 
 * Drop Python 3.4 (by :user:`John Kirkham <jakirkham>`; :issue:`89`).
 
+* Add codec :class:`GZip` to replace ``gzip`` alias for ``zlib``.
```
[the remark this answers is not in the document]
jakirkham: Good point. Went ahead and pushed the change to wrap things up here.

alimanfoo:
Thanks @jakirkham, merge at will.

alimanfoo:
Btw coverage is slightly down because of changes that came in from #93. I should have checked that, just needs a pragma no cover in the except block. Can include here or deal with separately, I don't mind.

jakirkham:
Thanks for pointing that out. Missed that. Decided to breakout the coverage fix in PR ( https://github.com/zarr-developers/numcodecs/pull/96 ) just to keep things easy to follow. HTH

jakirkham:
Thanks all. In it goes. 😄

PR #88, opened 2018-10-30
Fix README badges

jakirkham:
Some badges needed to be updated for the org. This makes those updates.

Skipped CI as it shouldn't be relevant for these changes.


alimanfoo:
Thanks for catching this.

PR #93, opened 2018-11-06
Run on lambda

tomwhite:
This is an addition to #83, which I needed to run on AWS lambda.

### DIFF
--- a/numcodecs/__init__.py
+++ b/numcodecs/__init__.py
@@ -44,7 +44,10 @@ try:
     from numcodecs.blosc import Blosc
     register_codec(Blosc)
     # initialize blosc
-    ncores = multiprocessing.cpu_count()
+    try:
+        ncores = multiprocessing.cpu_count()
+    except OSError:
+        ncores = 1
     blosc.init()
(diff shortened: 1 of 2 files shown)


alimanfoo:
Apologies @tomwhite, you did mention this, I forgot. Thanks for submitting.

tomwhite:
Thanks @alimanfoo.

numcodecs/__init__.py
```diff
@@ -44,7 +44,10 @@
     from numcodecs.blosc import Blosc
     register_codec(Blosc)
     # initialize blosc
-    ncores = multiprocessing.cpu_count()
+    try:
+        ncores = multiprocessing.cpu_count()
+    except OSError:
+        ncores = 1
```
jakirkham: Adding for reference: We need a `# pragma: no cover` for this block below since we don't have a way to test it really. PR ( https://github.com/zarr-developers/numcodecs/pull/96 ) addresses that.

tomwhite: Ah, didn't know about that. Thanks for reporting and fixing it @jakirkham!

PR #94, opened 2018-11-06
Refresh and rework requirements

alimanfoo:
This PR reworks the requirements files for easier maintenance:

* All indirect dependencies have been removed from requirements files.
* File requirements_dev.txt contains only direct requirements needed to run numcodecs, with versions pinned. This is something we would track with a bot to automatically upgrade.
* New file requirements_test.txt contains packages needed to run tests, not pinned. I have not pinned these because otherwise I fear the churn from a bot would get too much. 

TODO:


.travis.yml
```diff
@@ -22,7 +22,6 @@ matrix:
 
 install:
   - pip install -U tox-travis coveralls pip setuptools wheel
-  - pip install -r requirements_dev.txt
```
alimanfoo: This is not actually needed as installation is handled within tox.

numcodecs/tests/common.py
```diff
@@ -12,7 +12,7 @@
 
 
 from numcodecs.compat import buffer_tobytes, ndarray_from_buffer
-from numcodecs import *  # flake8: noqa
+from numcodecs.registry import get_codec
```
jakirkham: Think this might be related to the CI failure. Not sure though.

jakirkham:
Generally seems reasonable to me.

numcodecs/tests/common.py
alimanfoo: Yep, my bad. Should be fixed now.

alimanfoo:
Thanks @jakirkham. I'll wait for appveyor to pass then merge and setup PyUP if no objections.

jakirkham:
As long as it's green, I'm happy. 😄

PR #98, opened 2018-11-06
Fix msgpack issues and warnings

### LINKED ISSUE #85: msgpack: PendingDeprecationWarning: encoding is deprecated, Use raw=False instead
Using msgpack-python 0.5.6 I get:

```
/home/aliman/github/alimanfoo/zarr/.tox/py27/local/lib/python2.7/site-packages/numcodecs/msgpacks.py:48: PendingDeprecationWarning: encoding is deprecated, Use raw=False instead.
  items = msgpack.unpackb(buf, encoding=self.encoding)
```

alimanfoo:
This PR addresses a couple of issues relating to the new msgpack codec (codec ID "msgpack2") being introduced in numcodecs version 0.6.0:

* The ``encoding`` argument to ``packb`` and ``unpackb`` is being deprecated and is generating a ``PendingDeprecationWarning``. The arguments have been changed as recommended to use ``raw=False`` instead, with ``raw=True`` being available as an option. Two other relevant options, ``use_bin_type`` and ``use_single_float`` have also been surfaced from msgpack. Resolves #85.
* Round-trip encoding/decoding of an array of Python bytes objects was broken in the legacy msgpack codec, and works correctly in the new msgpack2 codec with new arguments set to defaults. Some tests have been added to verify the behaviour of old and new codecs with respect to both bytes and str (unicode) objects. This resolves an issue in zarr (https://github.com/zarr-developers/zarr/issues/324) where the tests currently expect broken behaviour from msgpack for an array of bytes objects.

N.B., because default values for some arguments being passed through to msgpack in the newer codec have changed in this PR, the fixture for the "msgpack2" codec has been regenerated. This is OK because the codec has not actually been released yet.

For reference, the initial work to add the new msgpack2 codec was done in PR #77.

TODO:


alimanfoo:
Any objections @zarr-developers/core-devs?

alimanfoo:
Thanks @jakirkham.

PR #119, opened 2018-11-08
Use old-style buffer on Python 2

jakirkham:
Makes use of the old-style buffer on Python 2 to more easily coerce objects to `memoryview`s on Python 2. The code here is a bit cleaner by smoothing over Python 2/3 differences. That said, it results in a ~100ns slowdown for Python 3 (due to the `lambda`) and a ~400ns slowdown for Python 2. This is pretty negligible on the grand scheme of things, but it is worth noting. Figured this might be worth discussion.

TODO:

### DIFF
--- a/numcodecs/compat.py
+++ b/numcodecs/compat.py
@@ -18,6 +18,7 @@ if PY2:  # pragma: py3 no cover
     binary_type = str
     integer_types = (int, long)
     reduce = reduce
+    buffer = buffer
 
 else:  # pragma: py2 no cover
 
@@ -25,6 +26,7 @@ else:  # pragma: py2 no cover
     binary_type = bytes
     integer_types = int,
     from functools import reduce
+    buffer = lambda a: a
 
 
 def buffer_tobytes(v):
@@ -33,10 +35,8 @@ def buffer_tobytes(v):
         return v
     elif isinstance(v, np.ndarray):
         return v.tobytes(order='A')
-    elif PY2:  # pragma: py3 no cover
-        v = buffer(v)
-
-    return memoryview(v).tobytes()
+    else:
+        return memoryview(buffer(v)).tobytes()
 
 
 def buffer_copy(buf, out=None):
--- a/numcodecs/compat.py
+++ b/numcodecs/compat.py
@@ -18,7 +18,6 @@ if PY2:  # pragma: py3 no cover
     binary_type = str
     integer_types = (int, long)
     reduce = reduce
-    buffer = buffer
 
 else:  # pragma: py2 no cover
 
@@ -26,7 +25,6 @@ else:  # pragma: py2 no cover
     binary_type = bytes
     integer_types = int,
     from functools import reduce
-    buffer = lambda a: a
 
 
 def buffer_tobytes(v):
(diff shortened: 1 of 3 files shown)


jakirkham:
> That said, it results in a ~100ns slowdown for Python 3 (due to the `lambda`)...

As a minor note, we eliminated a conditional branch that would have been ~25ns (not factored into the previous benchmark), so this is more accurately ~75ns slowdown for Python 3.

> ...and a ~400ns slowdown for Python 2.

Similarly we did not include the `isinstance` check that was here previously. When combined with the conditional branch, this is ~160ns. So the slowdown here is more accurately ~240ns for Python 2.

alimanfoo:
Thanks @jakirkham, didn't know about this.

FWIW I wouldn't be too concerned about performance of ``buffer_tobytes``, if code is having to go down that path then it usually implies there will be a memory copy (because of creating a new bytes object) so smaller timing changes probably won't have a noticeable impact. I.e., it is already a sub-optimal path for working around functions that don't support any buffer protocol.

I'm being picky but I'd be slightly reluctant to introduce ``buffer = lambda a: a`` in PY3, it makes it slightly obscure that there is PY2/3 compat going on within `buffer_tobytes``. I'd be happy to keep a conditional PY2 branch within ``buffer_tobytes`` to be more explicit, even if the extra conditional makes it slightly slower.

jakirkham:
Yeah, it's not well known. Just one of the subtleties of how the new buffer interface got added to Python 2 (much like the fact that Python 2's `array.array` doesn't support it 😄).

Sure. Not actually concerned personally. Just trying to be transparent is all. Was using largish (100s to 1000s) of elements to make sure there wasn't a copy. 

Completely understandable. Actually thought you might say that. Have reverted the last commit, which leaves a Python 2 branch in there. Technically this makes Python 3 slightly faster (though I don't want to micro-optimize 😉).

jakirkham:
So one thing that is a little concerning is the `tobytes` method of `memoryview`. This seems fine for most sizes of arrays, but if the array gets really large this starts to take a startling amount of time, which makes me wonder if there is a copy happening. Benchmarks taken on Python 3.

```python
In [1]: for i in range(9): 
    ...:     n = 10 ** i 
    ...:     r = 50 
    ...:     t = timeit.timeit( 
    ...:         "memoryview(a).tobytes()", 
    ...:         "import array; a = array.array('d', {n} * [1.0]);".format(n=n),
    ...:  
    ...:         number=r 
    ...:     ) 
    ...:  
    ...:     print("Took %f secs for %i elements." % (t / float(r), n))         
Took 0.000000 secs for 1 elements.
Took 0.000001 secs for 10 elements.
Took 0.000001 secs for 100 elements.
Took 0.000001 secs for 1000 elements.
Took 0.000004 secs for 10000 elements.
Took 0.000065 secs for 100000 elements.
Took 0.001999 secs for 1000000 elements.
Took 0.029316 secs for 10000000 elements.
Took 0.628060 secs for 100000000 elements.
```

alimanfoo:
I expect that tobytes() will make a memory copy. IIUC a Python bytes needs
to own its own memory, it can't be shared with another object. Hence we
only resort to creating a bytes object if we know the underlying
compression library cannot receive anything else. Normally we hope that the
underlying compression library can accept any object exposing the new-style
buffer interface, in which case we can pass objects straight through and
don't have to create bytes objects or do anything else that might imply a
memory copy. Hope that makes sense.

jakirkham:
Guess that makes sense. Would think that if the data was already read-only (e.g. another `bytes` object) then it would just reuse the data. Though that doesn't seem to be the case either. In any event, if this is expected, then I won't worry about.

jakirkham:
Have reworked this a bit in PR ( https://github.com/zarr-developers/numcodecs/pull/121 ), which seems like a nicer approach. So will close this out.

PR #121, opened 2018-11-09
Refactor and use more buffer view coercion

jakirkham:
This is a reworking of PR ( https://github.com/zarr-developers/numcodecs/pull/119 ). The main addition is the `to_buffer` function, which tries as best as possible to coerce input data into flat, contiguous, ~~unsigned character~~ views of the original data (without copying). This tries to follow the intent of the original `buffer_tobytes` implementation as much as possible. As a result `buffer_tobytes` is rewritten to be a thin wrapper around `to_buffer`.

Also tries to play to the strengths of Python 2 and Python 3 without too much special casing in other parts of the code. This amounts to returning `buffer` on Python 2 and `memoryview` on Python 3. As `memoryview` support is rather weak in Python 2, using `buffer` there ends up being a good choice. Besides creating a `memoryview` from a `buffer` in Python 2 is trivial. So this option is always available when needed (the reverse is not true however).

Updates the `BZ2`, `GZip`, `LZMA` and `Zlib` codecs to utilize `to_buffer` directly. This avoids the need for `tobytes` calls and `buffer_tobytes` in a few places. Dropping these copies should aid performance in these codecs.

Further this supplants the need for `handle_datetime` as views of data are always made to be `unsigned char` in all cases. Hence `handle_datetime` is dropped throughout. Also various checks for `object` type arrays are dropped throughout as this is also checked and handled in `to_buffer`. Generally consolidates the special casing in `to_buffer` allow the codecs to succinctly and clearly describe the main steps. Hopefully makes it easier for others to get started adding their own codecs (not that it was difficult before).

Adds tests of `to_buffer` alongside tests of `buffer_tobytes`. Includes some tests cases for expected errors to ensure they are raised.

TODO:


jakirkham:
Thanks for the feedback.

Yeah, hadn't meant to optimize Python 2 necessarily, but hopefully the code simplification is appreciated nonetheless. Would expect some performance bump on Python 3 as well. There are some other benefits outlined below.

So here's the really cool thing about `object` arrays. Basically a `memoryview` cast to `unsigned char` type and flattened has the same layout as calling `tobytes` on the `object` array, but provides a view instead. Example below on Python 3. On Python 2, casting and flattening are not options with `memoryview`s in Python. However the `buffer` type has the same effect (additionally marking the data as read-only, which we could extend to Python 3).

```python
In [1]: import numpy as np

In [2]: a = np.array(["22", 2+0j, 1, 1.3], dtype=object)

In [3]: np.asarray(memoryview(a.tobytes()))
Out[3]: 
array([160,  64, 228,   6,   1,   0,   0,   0, 240, 112,  83,  19,   1,
         0,   0,   0,  80,  72, 143,   5,   1,   0,   0,   0,  32, 237,
        28,  19,   1,   0,   0,   0], dtype=uint8)

In [4]: np.asarray(memoryview(a).cast('B'))
Out[4]: 
array([160,  64, 228,   6,   1,   0,   0,   0, 240, 112,  83,  19,   1,
         0,   0,   0,  80,  72, 143,   5,   1,   0,   0,   0,  32, 237,
        28,  19,   1,   0,   0,   0], dtype=uint8)
```

The advantage here is that we are able to get a view equivalent to calling `tobytes` on the `object` array, but without copying to a `bytes` instance yet. This of course extends to all other types as well. If the copy happens anyways, we simply make sure not to cause it any earlier. However there is the potential to exploit this view for higher performance.

That's a good point regarding typing. If we think `to_buffer` might be useful in other cases, would be happy to soften this `unsigned char` constraint. At least for these CPython included compressors (going off of the Python 3 implementations), they honor the buffer protocol for getting input data (a nice benefit of skipping a `tobytes` call 😉). However once they get the `Py_buffer` object they just act on the data as `char*` and seem to ignore the `itemsize` and `format` attributes. So it shouldn't matter for the cases changed here, but am happy to look more closely if you think it will be useful for other use cases.

alimanfoo:
For object arrays, if you take a buffer or convert to bytes, IIUC what you
are getting is access to the Python object memory addresses, i.e.,
pointers. If you are casting to unsigned char you're just breaking up those
pointers. Generally there's nothing good or useful you can do with those
pointers, which is why errors are raised when compression codecs are passed
an object array, and why zarr tries to ensure one of the codecs that
specifically serialises object arrays (like JSON, msgpack or VlenUTF8) is
called first.

Yes you're right most compressors will just cast to unsigned char and
process the memory as bytes. But some compressors and filters will want to
know itemsize where possible. So I think we should view as a different
dtype only where we know we have to to make things work, and otherwise
propagate unchanged as much information as we can via the buffer interface.

jakirkham:
Have relaxed the casting in the latest commit. Please let me know what you think.

jakirkham:
As to the object array case, think you are right that this is what is happening under the hood. That said, I must still be missing something. The same codecs handling various `object` arrays (e.g. Categorize, JSON, MsgPack, Pickle, Vlen*) are calling `buffer_tobytes`, which has been using `tobytes` on those NumPy arrays (and we are reproducing identical results with `to_buffer`). IOW do these codecs have bugs? At least a few of them appear to be using some sort of legacy implementation, in which case we know they were broken (hence deprecated and being replaced), but have not verified they are all like this. In any event, if we raise for `object` type arrays in `to_buffer`, we just cause some tests for these codecs to fail.

alimanfoo:
IIRC those codecs only use buffer_tobytes on decode, i.e., to get from the
serialised data back to an object array. Going in the other direction
(encode) they do something different.

numcodecs/tests/common.py
```diff
@@ -187,7 +187,8 @@ def check_backwards_compatibility(codec_id, arrays, codecs, precision=None, pref
         # setup
         i = int(arr_fn.split('.')[-2])
         arr = np.load(arr_fn)
-        arr_bytes = buffer_tobytes(arr)
+        if arr.dtype.kind is not 'O':
+            arr_bytes = buffer_tobytes(arr)
```
jakirkham: Have pulled this out separately in PR ( https://github.com/zarr-developers/numcodecs/pull/122 ).

numcodecs/compat.py
```diff
@@ -14,29 +14,47 @@
...
 
+def to_buffer(v):
+    """Obtain a `buffer` or `memoryview` for `v`."""
+    b = v
+    if isinstance(b, np.ndarray):
+        if b.dtype.kind is 'O':
+            raise ValueError('cannot encode object array')
+        b = b.reshape(-1, order='A')
+        if b.dtype.kind in 'Mm':
+            b = b.view(np.uint8)
```
alimanfoo: Minor point, but suggest to view this as int64, so itemsize is preserved.

numcodecs/compat.py
```diff
@@ -14,29 +14,47 @@
...
+        b = buffer(b)
+    elif PY2:  # pragma: py3 no cover
+        if not isinstance(b, array.array) and memoryview(b).format is 'O':
+            raise ValueError('cannot encode object array')
+        b = buffer(b)
+    else:      # pragma: py2 no cover
+        b = buffer(b)
+        if b.format is 'O':
+            raise ValueError('cannot encode object array')
+        b = b.cast('B').cast(b.format)
```
alimanfoo: Do we need to cast here? What does this add?

numcodecs/bz2.py
```diff
@@ -28,29 +27,21 @@ def __init__(self, level=1):
...
         if isinstance(buf, np.ndarray):
 
             # cannot compress object array
             if buf.dtype == object:
                 raise ValueError('cannot encode object array')
 
-            # if numpy array, can only handle C contiguous directly
-            if not buf.flags.c_contiguous:
-                buf = buf.tobytes(order='A')
+        buf = to_buffer(buf)
```
alimanfoo: Love how this is simplifying the codec implementations!

numcodecs/tests/test_compat.py
```diff
@@ -2,13 +2,22 @@
...
 
 from numcodecs.compat import buffer_tobytes
 
 
+def test_buffer_tobytes_raises():
+    a = np.array([u'Xin chào thế giới'], dtype=object)
+    with pytest.raises(ValueError):
+        buffer_tobytes(a)
+    with pytest.raises(ValueError):
+        buffer_tobytes(memoryview(a))
```
alimanfoo: Nice to test this, didn't realise you could get a memoryview of an object array.

numcodecs/compat.py
jakirkham: Thanks for catching this. Fixed.

numcodecs/compat.py
jakirkham: It flattens the data while preserving the type. This [SO answer]( https://stackoverflow.com/a/44486410 ) explains it pretty well.

Ideally we would just do `b.cast(b.format)`, but `memoryview`s seem to want to either start or end in some kind of `char` type unfortunately. So we pass through `unsigned char`.

numcodecs/tests/test_compat.py
jakirkham: Yeah, not sure how useful `object` type `memoryview`s are, but they do seem to be allowed.

numcodecs/bz2.py
jakirkham: Definitely. Wonder if it's worth extending this to the Cython implementations?

alimanfoo: IIRC the Cython implementations should already be using buffers in both Py2 and Py3, there is some C-level compatibility code in compat_ext.pyx. But by all means take a look, I may well have missed something!

alimanfoo:
Do we need to ensure buffers are read-only?

numcodecs/bz2.py
jakirkham: Did look briefly. Might save that for another PR though.

jakirkham:
Is it necessary? Maybe not. That said, `to_buffer` is always being used on input data that shouldn't be changed thus far. Also `buffer` in Python 2 always returns read-only buffers (though one could change this through the C API). So this just made the Python 2 and Python 3 behaviors a bit more similar.

numcodecs/compat.py
```diff
@@ -14,29 +14,48 @@
...
+        if b.dtype.kind in 'Mm':
+            b = b.view(np.uint64)
+
+    try:
+        b = memoryview(b)
+    except TypeError:  # pragma: py3 no cover
+        pass
+    else:
+        if b.format is not 'O':
+            b = np.array(b, copy=False).reshape(-1, order='A')
```
alimanfoo: I think the intention here is to ensure the buffer is read-only, but I'm afraid doing it this way will introduce some unintended memory copies. You could use np.frombuffer() instead which should guarantee no memory copy, but then you'd have to provide a dtype IIRC, and I think we want to interfere as little as possible with type information being passed through the buffer interface. Hence I was wondering if we can just do without this, i.e., pass the buffer/memoryview through and don't worry about whether it's read only or not.

jakirkham: Actually `np.array(b, copy=False)` does not copy `memoryview`s or `buffer`s. It takes a view. Have verified this in experiments on both Python 2 and 3. This has no effect on whether the buffer is readable or not (though it will inherit whatever setting the underlying buffer has).

jakirkham: Should add the main goal of this operation is to ensure that the data is flattened since many compressors expect this (and were doing that before).

jakirkham: Should note have included a test using NumPy's [`shares_memory`]( https://docs.scipy.org/doc/numpy-1.15.0/reference/generated/numpy.shares_memory.html ) to demonstrate this, which hopefully alleviates this concern.

jakirkham:
After giving it some thought, decided it was better if `to_buffer` return an `ndarray`. The reason being the `ndarray` supports both the `buffer` interface and the `memoryview` interface. As a result, a compressor can choose which interface it works best with and use it. Since we are able to construct an `ndarray` that views the original data, this ends up being a nice no-copy solution.

As a result have rewritten `to_buffer` to coerce non-`ndarray`s into either a `memoryview` or barring that a `buffer` (only on Python 2). Then it constructs an `ndarray` view of that data. With `memoryview`s, NumPy is able to constructing an array easily even preserving the shape, type, strides, etc.. Python 2's `buffer` simply provides less information; so, we have to either trust it has the right type, which works great for `mmap` (as it is consistent with a `memoryview` of `mmap` on Python 3), or we the tweak `ndarray` a little bit, which is what we do with the `array` case by casting back to the expected type. From there we do the same type checks and reshaping per usual on the `ndarray` and return the `ndarray` to the user.

jakirkham:
Did a little more tidying.

Now `ndarray_from_buffer` uses this new strategy of creating `ndarray` views onto the data. Moved type casting and reshaping out of `ndarray_from_buffer` to keep it general and ensure it always returns views onto the original data.

Rewrote `buffer_copy` to use `ndarray_from_buffer`, which simplifies its data handling. Also the `to_buffer` function calls `ndarray_from_buffer` and does some light type checking and reshaping. Finally the `buffer_tobytes` function has been dropped as it is trivial to call `tobytes` on the result of `to_buffer`.

Think this is ready for a thorough review at this stage. Also open to recommendations for a better name for `to_buffer` and any other feedback for that matter.

jakirkham:
Any thoughts on this @zarr-developers/core-devs?

jakirkham:
Thanks @alimanfoo. Appreciate the feedback.

Have to say I'm sorry this PR has become harder to understand. This is my fault as it started addressing a small point, but began to sprawl as we saw further simplifications could be made. Not to mention as work went into this it became clear there was a better direction we could go.

Think the most important change here is `ndarray_from_buffer`. That function has become a general purpose buffer accepting function that will return an `ndarray` from any kind of buffer on Python 2/3. It tries as best as possible to return an `ndarray` that matches the shape and type of the original buffer. As such many of the existing compat functions have either been dropped or largely replaced by calls to it.

However there are a number of situations that based on our discussion we wanted to gather into one place. For example we wanted to check for `object` type in one place. Also we had to handle the datetime/timedelta workaround. Finally some codecs expect a flattened array. So this logic has fallen into `to_buffer`. The name of the function is not great. The matter is complicated further as not every use case in Numcodecs needs or can even have these checks. Am half tempted to move this stuff back to the codecs as it is relatively light and it appears having it one place is confusing to others. Though am open to suggestions on how to improve this.

As to the Python 2/3 point, I think the confusion of utility functions is muddling this point. In actuality the compat code has shrunk quite significantly. There only is compat code in the `compat.py` module and its associated test suite (as opposed to throughout codecs as was the case before). In fact nearly all of it lives in `ndarray_from_buffer` and has `pragma: ... no cover` lines and/or `PY2` checks for the relevant Python version, which generally improves usability and readability elsewhere. IMHO this is one of the biggest wins and was part of the objective of this PR (the other being to cutdown on copies). Have tried to push a couple minor commits to highlight this further. Not sure how to improve this more, but open to suggestions.

Hope that helps provide some bearings to this PR. Sorry again that this PR has become overgrown. Would just like to stress this PR is trying to do 2 things consolidate handling of buffers, Python 2/3 differences, etc. by turning them all into `ndarray`s and cutting down on copies of the data.

alimanfoo:
Thanks @jakirkham, no problem at all. This is a really hard one to get right, there are so many moving parts - PY2 quirks, numpy quirks, compressor quirks. Very difficult to find a clean solution that accommodates all of these in an easy-to-digest way.

FWIW I think the key thing is that a codec should be able to express what type of input it requires, in a completely clear and obvious way. E.g., if a codec needs an object exporting C-contiguous memory via the new-style buffer interface, there should be an appropriately-named function for handling that. On the other hand, if a codec just needs a numpy array, but doesn't care if it's C-contiguous, there should a function for that.

jakirkham:
Yeah, that makes sense.

The other tricky part in my mind is finding the right number of utility functions. Certainly the naming in this PR needs work.

FWIW I wound up rolling a small library called [`cybuffer`]( https://github.com/jakirkham/cybuffer ) that takes anything that supports either buffer protocol and creates an object that exports `buffer` and `memoryview` compatible interfaces and works on Python 2/3. It is written in Cython so we could use this to solve similar issues there and is accessible from Python. IDK if we want to use it or not, but it did help learn more about the problem.

alimanfoo:
FWIW re naming, in #128 I've tentatively gone with ``ensure_ndarray()`` to mean make sure I've got a numpy array; and ``ensure_contiguous_ndarray()`` to mean make sure I've got a numpy array that can export C contiguous memory via the new-style buffer interface. In both cases, no memory copies will be made under any circumstances. 

Re cybuffer, it looks fierce and I'm sure I will be an invaluable resource, but for numcodecs I think we are doing OK with the approach we have in here / #128?

jakirkham:
Right, the naming point was more lamenting the poor naming choices in this PR. Agree that what has been added in PR ( https://github.com/zarr-developers/numcodecs/pull/128 ) is fine.

After the release it may be worth revisiting after whether we want to pull content from `cybuffer` or perhaps use it directly. In particular the bulk of the content in PR ( https://github.com/zarr-developers/numcodecs/pull/128 )'s `ensure_ndarray` can be shoved into `Buffer`, which would like a simplified version of what is in [`cybuffer`'s `__cinit__`]( https://github.com/jakirkham/cybuffer/blob/v0.2.0/src/cybuffer.pyx#L151 ).

alimanfoo:
Also just wanted to add that thanks primarily to work in this PR I think numcodecs is going to be in a **much** better place than it was before. I think it will also be a much better foundation for others wanting to contribute more codecs. Beers all round :beers:.

PR #122, opened 2018-11-09
Use `buffer_tobytes` on non-`object` types in test

jakirkham:
Pulled out from PR ( https://github.com/zarr-developers/numcodecs/pull/121 )

When running the `check_backwards_compatibility` test, make sure not to call `buffer_tobytes` on `object` arrays from fixture data. Casting these to bytes doesn't really make sense as it will return representation of pointers in the underlying buffer. Not only are the pointers things we do not want to be comparing, but it is also internal spec to NumPy and Python that really isn't reliable either. So make sure to pass through the `object` arrays as is.

[Description of PR]

TODO:

### DIFF
--- a/numcodecs/tests/common.py
+++ b/numcodecs/tests/common.py
@@ -187,8 +187,6 @@ def check_backwards_compatibility(codec_id, arrays, codecs, precision=None, pref
         # setup
         i = int(arr_fn.split('.')[-2])
         arr = np.load(arr_fn)
-        if arr.dtype.kind is not 'O':
-            arr_bytes = buffer_tobytes(arr)
         if arr.flags.f_contiguous:
             order = 'F'
         else:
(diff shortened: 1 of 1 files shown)


numcodecs/tests/common.py
```diff
@@ -187,7 +187,8 @@ def check_backwards_compatibility(codec_id, arrays, codecs, precision=None, pref
         # setup
         i = int(arr_fn.split('.')[-2])
         arr = np.load(arr_fn)
-        arr_bytes = buffer_tobytes(arr)
+        if arr.dtype.kind is not 'O':
+            arr_bytes = buffer_tobytes(arr)
```
alimanfoo: Being picky, but wonder if simplest thing to do here is move this down to the point at which the comparison is made. I.e., line 236 below:

```
assert arr_bytes == buffer_tobytes(dec)
```

...becomes:

```
assert buffer_tobytes(arr) == buffer_tobytes(dec)
```

jakirkham:
Thanks for the review. :)

PR #127, opened 2018-11-21
Cast datetime and timedelta to signed 64-bit int

jakirkham:
Related to issue ( https://github.com/zarr-developers/zarr/issues/342 )

The `NaT` type is represented as `-0`. As a result, casting to an unsigned integral fails and throws an error. However casting to a signed integral type does not have this problem and proceeds without issues.

TODO:


alimanfoo:
Hi @jakirkham, this looks good, but there is an issue to do with how the backwards compatibility tests are setup, I'll try to explain. Basically the list of arrays in the test module for each codec can be added to, but existing arrays in that list shouldn't be changed. If they are changed, then they will be out of sync with the fixture data. This won't show up in the tests, as previous test arrays get saved to files, so the backwards compatibility tests run off the previously saved data.

In other words, if you change any of the arrays in the list, then the corresponding fixture data should also be deleted and regenerated. You can find the data to delete because the order in which the arrays appear in the list is used to generate a file name. E.g., within the test_zlib module, the first array in the list gets saved as [fixture/zlib/array.00.npy](https://github.com/zarr-developers/numcodecs/blob/master/fixture/zlib/array.00.npy), and the corresponding encoded data from the first codec in the codecs list gets saved as [fixture/zlib/codec.00/encoded.00.dat](https://github.com/zarr-developers/numcodecs/blob/master/fixture/zlib/codec.00/encoded.00.dat). I.e., if you changed the first array in the list, then you should delete ``fixture/zlib/array.00.npy`` and all files like ``fixture/zlib/*/encoded.00.dat``.

Alternatively and probably better in this case, leave all the existing arrays as-is, and append some new arrays to the list, in which case some new fixture data will get generated on the first test run.

Hope that makes sense!

jakirkham:
Thanks @alimanfoo. IIUC you would like the tests added as a new set in the parameterized group. If so that makes sense to me, sorry I didn't think about it earlier. Have added the `NaT` tests as new tests at the end of the parameterized tests, generated the new fixture data locally (via running `pytest`), and committed/pushed. Please let me know if that works or if more needs to be done.

numcodecs/tests/test_blosc.py
```diff
@@ -47,6 +47,10 @@
     np.random.randint(0, 2**60, size=1000, dtype='u8').view('m8[ns]'),
     np.random.randint(0, 2**25, size=1000, dtype='u8').view('M8[m]'),
     np.random.randint(0, 2**25, size=1000, dtype='u8').view('m8[m]'),
+    (-np.random.randint(-2**30, 2**30, size=1000, dtype='i8')).view('M8[ns]'),
```
alimanfoo: Out of interest, if this is random numbers between -2**30 and 2**30, why do we need to take the negative of the array? Is it not enough to specify the limits within dancing?

alimanfoo:
Going to merge this one so we can rebase other PRs (#121, #128).

PR #128, opened 2018-11-22
Use more buffers (redux)

alimanfoo:
Based on work in #121, trying to make a clean split between PY2 and PY3 code paths, and make function naming more intuitive.

TODO:


numcodecs/compat.py
```diff
@@ -27,80 +25,150 @@
...
-    if isinstance(l, text_type):
-        return l
-    else:  # pragma: py3 no cover
-        return text_type(l, encoding=encoding)
+        if PY2:  # pragma: py3 no cover
+            # accept objects exposing either old-style or new-style buffer interface
+            try:
+                o = memoryview(o)
+            except TypeError:
+                o = buffer(o)
```
jakirkham: Would use `np.getbuffer` here. This behaves the same as `buffer` except it will preserve the same write permissions on the underlying data; whereas, `buffer` forces this to be read-only regardless.

numcodecs/compat.py
```diff
@@ -27,80 +25,150 @@
...
 
+        # first try to obtain a memoryview or buffer, needed to ensure that we don't
+        # subsequently copy memory when going via np.array()
 
-def ensure_text(l, encoding='utf-8'):
-    if isinstance(l, text_type):
-        return l
-    else:  # pragma: py3 no cover
-        return text_type(l, encoding=encoding)
+        if PY2:  # pragma: py3 no cover
```
jakirkham: What's the advantage of the `if` over just the `try...catch`?

numcodecs/compat.py
```diff
@@ -27,80 +25,150 @@
...
+        o = np.array(o, copy=False)
+
+    # check for object arrays
+    if o.dtype == object:
+        raise ValueError('object arrays are not supported')
+
+    if flatten:
+
+        # flatten the array to 1 dimension
+        o = o.reshape(-1, order='A')
```
jakirkham: Also thought about this as well. Have mixed feelings about including an option for something that is this simple to do externally. Though can certainly see the value in including it for end users. Would be good to weigh the value of including it vs. not.

numcodecs/compat.py
```diff
@@ -27,80 +25,150 @@
...
+
+    return o
+
+
+def ensure_memoryview(o, flatten=True):
+    """Obtain a :class:`memoryview` with a view of memory exposed by `o`.
```
jakirkham: Originally went down this route as well. However what I discovered is that in Python 2 some things only support the old-style buffer protocol and some support the new buffer protocol or both. Admittedly Python 3 is much more straightforward as everything supports the new  buffer protocol where possible.

Given this realization it seemed much better to return something that supported both protocols and was generally more functional (to support reshaping, casting, type checking, etc.). Fortunately NumPy's `ndarray` does all of these things. So it just made more sense to always convert to that. Plus contributors are probably more comfortable with NumPy `ndarray`s than `memoryview`s.

numcodecs/compat.py
```diff
@@ -27,80 +25,150 @@
...
 
     if out is None:
         # no-op
         return buf
 
-    # handle ndarray destination
-    if isinstance(out, np.ndarray):
+    # obtain ndarrays, casting to the same data type
+    buf = ensure_ndarray_from_memory(buf).view('u1')
+    out = ensure_ndarray_from_memory(out).view('u1')
```
jakirkham: Why cast to `uint8`?

numcodecs/compat.py
```diff
@@ -27,80 +25,150 @@
...
+            o = memoryview(o)
+
+        # N.B., this is not documented, but np.array() will accept an object exposing
+        # a buffer interface, and will take a view of the memory rather than making a
+        # copy, preserving type information where present
+        o = np.array(o, copy=False)
+
+    # check for object arrays
+    if o.dtype == object:
+        raise ValueError('object arrays are not supported')
```
jakirkham: This gets a bit tricky in practice. Not all things we want to use this with can enforce that it is `object` only. `Categorize` is a prime example. It'd be nice if we could separate this somehow so we can make use of this strategy more often.

numcodecs/compat.py
```diff
@@ -27,80 +25,150 @@
...
+        -------
+        o : buffer
+
+        """
+
+        # go via numpy, for convenience
+        o = ensure_ndarray_from_memory(o)
 
+        # expose as buffer
+        o = buffer(o)
```
jakirkham: Already mentioned above, but would suggest using `np.getbuffer` as well to ensure write permissions are preserved.

Though much like `memoryview` I think it is easier to just always give an `ndarray` instead since it supports both protocols. That avoids getting into too much code that is different between Python 2/3 and is generally much more functional for end users.

This is just my opinion though. Feel free to disagree. Would be interested in hearing your thoughts on this as well.

numcodecs/compat.py
alimanfoo: FWIW I'm trying to create a clean separation for all PY2 code paths. The try...catch is only needed under PY2. So it will be obvious this whole conditional block can be deleted when PY2 is dropped.

numcodecs/compat.py
alimanfoo: Yep, could do. FWIW no codecs require a writable buffer, and it's not immediately obvious what np.getbuffer() is doing without doing some digging, so I had a slight preference for a bare ``buffer()`` call to make it more obvious this is accessing old-style buffers.

numcodecs/compat.py
alimanfoo: This is tricky. The way I've handled this currently is that this function is purely about obtaining a ndarray with contiguous memory, in the broader sense that all memory that needs to be encoded is in a single block (I've renamed the function to ``ensure_contiguous_ndarray()``). If a codec wants to be more flexible and accept object arrays also, then they can do that, but they need to handle that differently. Examples are ``Categorize``, ``MsgPack`` and ``JSON``.

numcodecs/compat.py
alimanfoo: FWIW the only codec that actually needs the ``ensure_buffer()`` function is ``ZLib``. 

Also FWIW the reason why I didn't just use ``ensure_contiguous_ndarray()`` everywhere, and get rid of ``ensure_memoryview()`` and ``ensure_buffer()``, is because of arrays with datetime or timedelta data types. Codecs like ``ZLib`` and ``BZ2`` and anything that will try to access data via the new-style buffer interface in Python 3 will error on an array with datetime or timedelta datatype.

I could push the conversion from datetime/timedelta to integer datatype up inside ``ensure_contiguous_ndarray()``, which would mean that that single function could be used everywhere, and both ``ensure_memoryview()`` and ``ensure_buffer()`` could be deleted. Seems a bit beyond the scope of just ensuring contiguous memory, but it would simplify the code.

numcodecs/compat.py
jakirkham: Is this being used by `memory_copy` though?

As to what NumPy is doing, It's a good question. Though it turns out to be very simple. Here's the [C code]( https://github.com/numpy/numpy/blob/de28edd8f514b82c0524b55f622078d47f479322/numpy/core/src/multiarray/multiarraymodule.c#L3493 ).

Basically it tries to get a writable pointer to the object's data, if it fails it creates a read-only buffer. If it succeeds, it creates a read-write buffer. That's pretty much it.

numcodecs/compat.py
jakirkham: Yeah I can see how there is a little confusion either way. If this is preferred, let's go with it.

numcodecs/compat.py
jakirkham: Yeah, I dealt with the same tension regarding the scope of the `ndarray` function. Though increasingly I think we should just move the datetime/timedelta code there and think of a better function name.

numcodecs/compat.py
alimanfoo: I've removed the option in latest commit.

numcodecs/compat.py
alimanfoo: I think it's needed to ensure source and destination line up? Could be wrong though.

alimanfoo:
Thanks @jakirkham for the comments, this is a tricky one to get right.

FWIW I think the key issue here is that many codecs require an input that, under Python 3, exposes a new-style buffer interface onto a C contiguous block of memory. E.g., things like ``ZLib``, ``BZ2`` and ``LZMA``. 

A blocker from just passing a numpy array to these functions is the datetime & timedelta datatypes which  you can't take a memoryview of. 

If I included a conversion from datetime or timedelta to int64 **within** the ``ensure_contiguous_ndarray()`` function, then it would be possible to use that function everywhere instead of either ``ensure_memoryview()`` or ``ensure_buffer()``, and so delete those two functions. Maybe that would be a good idea.

numcodecs/compat.py
jakirkham: My leaning would to write `ensure_memoryview(...)` as `memoryview(ensure_contiguous_ndarray(...))` where possible and try to make sure that works (changing `ensure_contiguous_ndarray` as needed). Expect that users are quite happy to stick with `ndarray`s and only go to a `memoryview` in unusual circumstances. Plus it will be lower cognitive load (even for us 😄).

numcodecs/compat.py
alimanfoo: Yep, maybe it is important for ``memory_copy``, good point.

Yeah ``np.getbuffer()`` is simple, but you do have to go trawling the numpy C source code for it. I've spent a fair amount of time today trawling numpy C source code to figure out WTF it's doing, especially inside ``np.array()``, was just thinking to save others the pain :smiley:.

jakirkham:
> If I included a conversion from datetime or timedelta to int64 **within** the `ensure_contiguous_ndarray()` function, then it would be possible to use that function everywhere instead of either `ensure_memoryview()` or `ensure_buffer()`, and so delete those two functions. Maybe that would be a good idea.

This is my temptation as well. Let's go for it. 😄

numcodecs/compat.py
alimanfoo: Yeah, maybe ``gimme_an_ndarray_with_a_C_contiguous_memory_buffer_that_supports_the_new_style_buffer_protocol_and_doesnt_have_an_object_dtype()`` :smiley:. Seriously though, that is the essence of the function, so some name that captures that would be good, and I'd be happy to move the datetime/timedelta code in.

numcodecs/compat.py
jakirkham: Yeah, I feel your pain. Same thing with the (new) buffer protocol support. It's probably worth writing up an issue for NumPy devs requesting clearer documentation about this stuff.

numcodecs/compat.py
jakirkham: Hmm...guess I'd be tempted not to do any reshaping in here. Mainly thinking of this for two reasons. First it can introduce a copy potentially. Second users may need to do other things that will cause flattening anyways (e.g. `tobytes`).

numcodecs/compat.py
jakirkham: Sure, casting may make sense. What about casting `buf` to match `out` though?

numcodecs/compat.py
jakirkham: Yeah, it starts to feel like we are writing Java code. 😄

Having `serializable` in there sounds really good for some reason. 😉

numcodecs/categorize.py
```diff
@@ -53,7 +53,10 @@ def __init__(self, labels, dtype, astype='u1'):
     def encode(self, buf):
 
         # view input as ndarray
-        arr = ndarray_from_buffer(buf, self.dtype)
+        if self.dtype == object:
+            arr = np.asanyarray(buf, dtype=object).reshape(-1, order='A')
```
jakirkham: Just out of curiosity, why is `asanyarray` needed?

numcodecs/checksum32.py
```diff
@@ -7,30 +7,28 @@
...
 
     def encode(self, buf):
-        if isinstance(buf, np.ndarray) and buf.dtype == object:
-            raise ValueError('cannot encode object array')
-        arr = ndarray_from_buffer(buf, dtype='u1')
+        arr = ensure_contiguous_ndarray(buf).view('u1')
         checksum = self.checksum(arr) & 0xffffffff
         enc = np.empty(arr.nbytes + 4, dtype='u1')
         enc[:4].view('<u4')[0] = checksum
         enc[4:] = arr
```
jakirkham: Should we use `memory_copy` here?

jakirkham:
> OK, 'tis done! I'm using ``ensure_contiguous_ndarray()`` everywhere now. ``ensure_memoryview()`` and ``ensure_buffer()`` are gone.

Beautiful! Thanks for doing that. 😄

> There were still some quirks in the ``BZ2`` and ``GZip`` codecs that required an explicit conversion to ``memoryview``, but I figure that's OK.

Yeah, I ran into these two as well. Agree this seems fine.

> Let me know what you think.

Added a few more comments above in our existing threads to keep continuity with the existing discussion. A couple other comments above as well. For the most part these are moving to minor points. IOW this is looking pretty good.

numcodecs/categorize.py
alimanfoo: I'm guessing that any of ``np.array(buf, dtype=object, copy=False``, ``np.asarray(buf, dtype=object)`` or ``np.asanyarray(buf, dtype=object)`` would have the desired effect of creating an array from buf, without copying if buf is already a numpy array with object dtype. Not sure which should be preferred.

numcodecs/checksum32.py
alimanfoo: Possibly could do, I don't know if would would be any faster.

numcodecs/compat.py
alimanfoo: Yes some sort of feedback to numpy devs would seem appropriate, not quite sure how to frame it yet though. One thing I find particularly confusing is the circumstances under which ``np.array()`` will use the buffer interface (and share memory) versus when it will not. E.g., if I have a bytes object like:

```
a = b'asdfghjkl'
```

...then the following **will** result in a copy:

```
b = np.array(a, copy=False)
```

...whereas the following **will not** result in a copy:

```
b = np.array(memoryview(a), copy=False)
```

The docstring for ``np.array()`` does not even mention the buffer interface at all, although it does mention the "array interface", but I thought that was a different thing, although I may be confused.

Also, numpy has the ``np.frombuffer()`` function, which you would think is exactly what we're after here, as it is explicitly documented as a way to obtain a numpy array from some object exposing a memory buffer. But in order to use that function, you have to provide a dtype explicitly. Numpy could (should?) be obtaining dtype information from the buffer interface by default IMO. You could potentially work around that by doing something like:

```
m = memoryview(a)
b = np.frombuffer(m, dtype=m.format)
```

...but then I'm not completely clear on whether it's cool to use the struct format codes as a numpy dtype. It seems to work, but I did also find an open issue about incompatibilities (https://github.com/numpy/numpy/issues/9642).

numcodecs/compat.py
alimanfoo: So I think this depends on the scope of the function, i.e., who needs it and why. 

Forget about PY2 for a minute, let's consider just PY3.

The compressor codecs ``ZLib``, ``BZ2`` and ``LZMA`` require an object that exports the new-style buffer interface (e.g., ``memoryview``) with C-contiguous memory. So for them, it is useful to reshape the input, because it ensures a C-contiguous array, whether the input is C-contiguous, F-contiguous or non-contiguous. IIUC if the input is either C- or F-contiguous then calling ``reshape(-1)`` **will not** result in a copy. If the input is non-contiguous then a copy may occur, but then this is unavoidable - the alternative would be to raise an error that the input needs to be contiguous.

However, other codecs may have less stringent requirements. E.g., they may need a numpy array, but may not care if it is contiguous.

So possibly we should have more than one convenience function for obtaining a numpy array, to account for these differences, with some sort of clear naming scheme, so it is clear which function you should use depending on what requirements you have.

numcodecs/categorize.py
jakirkham: It's a fair question. My thinking would be one of the first two. `np.asanyarray` seems to suggest we will preserve other `ndarray` subclasses that might include masks, axis tags, or other user customizations. When in reality, we are only serializing the array itself.

numcodecs/checksum32.py
jakirkham: Was mainly thinking about this from a consistency standpoint.

numcodecs/compat.py
```diff
@@ -27,80 +26,96 @@
...
+        arr = np.array(mem, copy=False)
 
-        # copy via numpy
-        np.copyto(out, buf)
+        if PY2 and isinstance(buf, array.array):  # pragma: py3 no cover
+
+            # type information will not have been propagated via the old-style buffer
+            # interface, so we have to manually hack it after the fact
+            if buf.typecode == 'u':
+                t = 'U1'
```
jakirkham: This one gets more complicated.

First off the unicode array type is [totally broken]( https://bugs.python.org/issue13072 ) and [deprecated]( https://github.com/python/cpython/commit/029780be09ab5483ac4562d795ea4b4b6d456f44 ). Second it maps to `Py_UNICODE`, which can be [build dependent]( https://bugs.python.org/issue13072#msg168005 ). On Windows this is UCS-2 because of [historic reasons]( https://en.wikipedia.org/wiki/Unicode_in_Microsoft_Windows ). On other platforms this is usually UCS-4 (with the exception of [Python 2.7 on macOS]( https://github.com/Homebrew/legacy-homebrew/pull/32368 ) mostly due to community inertia). Third `memoryview`s represent UCS-2 and UCS-4 as different types (i.e. checking `.format`) where the first is `"u"` and the second is `"w"`. Fourth NumPy only supports UCS-4 not UCS-2 as noted in [this issue]( https://github.com/numpy/numpy/issues/12369 ). Fortunately NumPy will give a better error message about this in NumPy 1.16. 

If we are to go down the road of supporting these text types (which also includes `"c"` on Python 2.7), my suggestion would be that we just cast them all to appropriate width unsigned integers, which interestingly has [come up before]( https://bugs.python.org/issue13072#msg167997 ). Possibly raising a warning along the way to point out these types will be dropped in Python 4. That way this data can always safely be represented by NumPy or anything else without needing to know anything about these deprecated or removed text types. Alternatively we could raise an error if users give us this sort of data and not support it.

alimanfoo: Argh, OK, now the code in #121 makes sense.

alimanfoo:
Hi @jakirkham, after some thrashing around I think I have basically converged on something very similar to what you have in #121. There are some small differences which I'm more than happy to discuss. I've also pushed on and refactored the buffer compatibility code in the cython modules, to re-use the new compat functions. I still have to address the point you raised about dealing with unicode arrays, but apart from that, I'd be very grateful if you could take a look and see if there's anything you think I've missed, or anything you think should be done differently.

numcodecs/compat.py
jakirkham: That is expected. Windows is using UCS-2 and NumPy only supports UCS-4. Casting is pretty much the only way around this AFAIK.

alimanfoo: OK, now I completely understand why there was array casting before and after conversion to numpy array. Thanks!

alimanfoo:
In the interests of getting to a 0.6 release asap, I'd like to suggest we move forward with this PR instead of #121, mainly for the reason that I've had the chance to add in a bit more documentation and comments, so hopefully when we or someone else needs to revisit this in future, they should be able to comprehend what is being done and why. I believe that the approaches taken here and in #121 are essentially the same, with the main differences being in function naming and some minor implementation details. Please let me know if any objections. Whatever PR we take forward, primary credit for this work should go to @jakirkham.

numcodecs/compat.py
```diff
@@ -27,80 +26,161 @@
...
-            buf = buf.reshape(out.shape, order=order)
+    """
 
-        # copy via numpy
-        np.copyto(out, buf)
+    if isinstance(buf, np.ndarray):
+        # already a numpy array
+        arr = buf
+
+    elif isinstance(buf, array.array) and buf.typecode == 'u':
```
jakirkham: Could we also do this for [`array`'s of type `"c"`]( https://docs.python.org/2/library/array.html#module-array )?

alimanfoo: I think arrays of type "c" actually work ok, come out as numpy type "S1". There's a test for this in test_compat.

numcodecs/compat.py
```diff
@@ -27,80 +26,161 @@
...
+def ensure_text(l, encoding='utf-8'):
+    if isinstance(l, text_type):
+        return l
+    else:  # pragma: py3 no cover
+        return text_type(l, encoding=encoding)
 
 
-def buffer_copy(buf, out=None):
-    """Copy the contents of the memory buffer from `buf` to `out`."""
+def ensure_ndarray(buf, dtype=None):
```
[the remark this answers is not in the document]
alimanfoo: I went back and forth on that, several times. In the end a lot of codecs wanted a view, and it looked cleaner to include as an argument to ``ensure_ndarray()``. Also parallels the dtype argument to numpy array creation functions, not that that's a big deal.

numcodecs/compat.py
```diff
@@ -27,80 +26,161 @@
...
+    if dst is None:
+        # no-op
+        return src
+
+    # ensure ndarrays
+    src = ensure_ndarray(src)
+    dst = ensure_ndarray(dst)
+
+    # ensure same data type
+    if dst.dtype != object:
```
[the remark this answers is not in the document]
alimanfoo: I can't remember if having ``dst`` with ``object`` type is specifically covered. Actually it's an area where coverage could be improved, what happens when different shapes and types of array are provided as the ``out`` argument to ``codec.decode()``. We could remove the check, in which case an error would get raised on any attempt to view as object dtype. I think I left the check because then it would be possible to copy data into an object array, although I'm not entirely sure that's something we want to support. I believe ``np.copyto()`` can handle object arrays in a sensible way.

numcodecs/gzip.py
```diff
@@ -28,22 +25,9 @@ def __init__(self, level=1):
...
-            # if numpy array, can only handle C contiguous directly
-            if not buf.flags.c_contiguous:
-                buf = buf.tobytes(order='A')
-
-        if PY2:  # pragma: py3 no cover
-            # ensure bytes, PY2 cannot handle things like bytearray
-            buf = buffer_tobytes(buf)
+        # normalise inputs
+        buf = ensure_contiguous_ndarray(buf)
+        buf = memoryview(buf)  # needed on PY2 for unknown reasons
```
jakirkham: Had run into this myself as well, but then started wondering why it was an issue.

Turns out discussion in [Python issue 7471]( https://bugs.python.org/issue7471 ) generated [a patch]( https://github.com/python/cpython/commit/673ddf9907898d4483a2622699f588f0b58a5226 ) that made `GzipFile` a subclass of `io.BufferedIOBase`. Part of the patch included an [explicit check for `memoryview` instances]( https://github.com/python/cpython/commit/673ddf9907898d4483a2622699f588f0b58a5226#diff-90c90c977f09602928f0518fc849ca9fR213 ), which [it converted to `bytes`]( https://github.com/python/cpython/commit/673ddf9907898d4483a2622699f588f0b58a5226#diff-90c90c977f09602928f0518fc849ca9fR214 ). Unfortunately there is not much detail as to why this conversion was needed in that discussion.

A follow-up in [Python issue 23688]( https://bugs.python.org/issue23688 ), which attempts to use `memoryview`s without converting to `bytes` [clarifies this point]( https://bugs.python.org/issue23688#msg238307 ). In particular, [`len`]( https://github.com/python/cpython/commit/673ddf9907898d4483a2622699f588f0b58a5226#diff-90c90c977f09602928f0518fc849ca9fR217 ) is used to determine the number of bytes in the data. However this result will be wrong if the data is not of `char` type. This was [fixed]( https://github.com/python/cpython/commit/bca63b362d23f154a5ed7fe43e4146977bba181e ) in Python 3.5+ to actually check `nbytes`, but [that functionality does not exist in Python 2.7]( https://bugs.python.org/issue23688#msg239034 ); so, it was not backported.

Though I think we can do better as we use NumPy. Namely we can change this line to `buf = buf.view('u1')` instead of converting to a `memoryview`, which should avoid the copy and get the correct behavior. We can limit this to Python 2 if that is preferred.

numcodecs/compat.py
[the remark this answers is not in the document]
jakirkham: Yeah I can see your point. Just am a little leary adding options to functions that are easy for users to perform afterwards. Though I don't have strong feelings about it if you would rather keep it.

jakirkham:
Had some pretty minor comments above. Think this is basically done at this stage.

> In the interests of getting to a 0.6 release asap, I'd like to suggest we move forward with this PR instead of #121...

Sounds good to me.

> ...mainly for the reason that I've had the chance to add in a bit more documentation and comments, so hopefully when we or someone else needs to revisit this in future, they should be able to comprehend what is being done and why.

I think the test coverage and function naming were also better here. The API has also improved generally.

> Whatever PR we take forward, primary credit for this work should go to @jakirkham.

Thanks for working on this as well. It has come a long way since that PR.

numcodecs/compat.py
jakirkham: Yeah, this wound up being cleaner to handle in Cython interestingly. 😆

`memoryview`'s `cast` is not nearly as functional as `ndarray`'s `view`.

numcodecs/compat.py
alimanfoo: Good point, I've disallowed "c" in latest commit.

numcodecs/compat.py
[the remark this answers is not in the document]
alimanfoo: I've gone with your preference in latest commit, I think it's good to be explicit that the codec is taking a view.

numcodecs/gzip.py
alimanfoo: You're write, viewing as 'u1' does work on PY2, made this change in latest commit.

jakirkham: Adding a link back to this from the code in PR ( https://github.com/zarr-developers/numcodecs/pull/133 ) lest we wonder what this workaround is doing. 😄

PR #130, opened 2018-11-23
Use setup_requires for cpuinfo

alimanfoo:
Use ``setup_requires`` to obtain ``cpuinfo`` module rather than vendoring it. (Hopefully) supersedes #82.

TODO:


.travis.yml
```diff
@@ -22,7 +22,6 @@ matrix:
 
 install:
   - pip install -U tox-travis coveralls pip setuptools wheel
-  - python setup.py build_ext --inplace
```
alimanfoo: This isn't actually needed, building will happen via tox.

jakirkham: Agreed. Think this was causing the build to happen twice before actually.

setup.py
```diff
@@ -23,9 +22,15 @@
 
 
 # determine CPU support for SSE2 and AVX2
-cpu_info = cpuinfo.get_cpu_info()
-have_sse2 = 'sse2' in cpu_info['flags']
-have_avx2 = 'avx2' in cpu_info['flags']
+try:
+    import cpuinfo
```
jakirkham: Unfortunately this actually won't benefit from `setup_requires` at this stage as Setuptools hasn't had the opportunity to run this. To get this to work correctly, we need to add a [`BuildExtCommand` subclass like this one for NumPy]( https://github.com/jakirkham/cookiecutter-cython/blob/de7397f77ba5118c9e5973966f925d2a4af220a7/%7B%7Bcookiecutter.project_name%7D%7D/setup.py#L21-L27 ) and add it to [our `cmdclasses`]( https://github.com/jakirkham/cookiecutter-cython/blob/de7397f77ba5118c9e5973966f925d2a4af220a7/%7B%7Bcookiecutter.project_name%7D%7D/setup.py#L73 ) in [`setup`]( https://github.com/jakirkham/cookiecutter-cython/blob/de7397f77ba5118c9e5973966f925d2a4af220a7/%7B%7Bcookiecutter.project_name%7D%7D/setup.py#L159 ).

alimanfoo: Doh, of course.

alimanfoo:
On reflection I think I'm going to push back work to avoid vendoring py-cpuinfo to the next release. It's not straightforward to use a ``BuildExtCommand`` subclass for the blosc extension. Better to wait for full support for pyproject.toml IMO.

PR #134, opened 2018-11-28
Cutdown copies with BytesIO

jakirkham:
This is currently an exploratory PR, which tries to use `BytesIO` to avoid some copies.

TODO:

### DIFF
--- a/numcodecs/gzip.py
+++ b/numcodecs/gzip.py
@@ -59,7 +59,7 @@ class GZip(Codec):
                 out = ensure_contiguous_ndarray(out)
                 nbytes = decompressor.readinto(out)
                 nbytes += len(decompressor.read())
-                if nbytes != out.nbytes:
+                if nbytes > out.nbytes:
                     raise ValueError("`out` unable to fit %i bytes" % nbytes)
             else:
                 out = ensure_ndarray(decompressor.read())
--- a/numcodecs/gzip.py
+++ b/numcodecs/gzip.py
@@ -5,7 +5,7 @@ import io
 
 
 from .abc import Codec
-from .compat import ndarray_copy, ensure_contiguous_ndarray, PY2
+from .compat import ensure_ndarray, ensure_contiguous_ndarray, PY2
 
 
 class GZip(Codec):
@@ -42,24 +42,26 @@ class GZip(Codec):
         try:
             compressed = compressed.getbuffer()
         except AttributeError:  # pragma: py3 no cover
-            compressed = memoryview(compressed.getvalue())
+            compressed = compressed.getvalue()
 
-        return np.array(compressed, copy=False)
+        return ensure_ndarray(compressed)
 
     # noinspection PyMethodMayBeStatic
     def decode(self, buf, out=None):
 
         # normalise inputs
         buf = ensure_contiguous_ndarray(buf)
-        if out is not None:
-            out = ensure_contiguous_ndarray(out)
 
         # do decompression
         buf = io.BytesIO(buf)
         with _gzip.GzipFile(fileobj=buf, mode='rb') as decompressor:
-            decompressed = decompressor.read()
+            if out is not None:
+                decompressed = ensure_contiguous_ndarray(out)
+                decompressor.readinto(decompressed.view('u1'))
+            else:
+                decompressed = ensure_ndarray(decompressor.read())
 
         # handle destination - Python standard library zlib module does not
         # support direct decompression into buffer, so we have to copy into
         # out if given
-        return ndarray_copy(decompressed, out)
+        return decompressed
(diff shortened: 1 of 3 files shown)


alimanfoo:
This looks nice :+1:

alimanfoo:
FWIW this looks good to go to me, and would be nice to include in 0.6.0.

numcodecs/gzip.py
```diff
@@ -38,24 +38,30 @@ def encode(self, buf):
...
-        if out is not None:
-            out = ensure_contiguous_ndarray(out)
 
         # do decompression
         buf = io.BytesIO(buf)
         with _gzip.GzipFile(fileobj=buf, mode='rb') as decompressor:
-            decompressed = decompressor.read()
+            if out is not None:
+                decompressed = ensure_contiguous_ndarray(out)
+                decompressor.readinto(decompressed)
```
jakirkham: Should we check that the buffer was exhausted here? If so, what should we raise if that does not happen?

numcodecs/tests/test_gzip.py
```diff
@@ -93,3 +93,22 @@ def test_err_encode_non_contiguous():
...
+def test_err_out_too_small():
+    # non-contiguous memory
+    arr = np.arange(10, dtype='i4')
+    out = np.empty_like(arr)[:-1]
+    for codec in codecs:
+        with pytest.raises(ValueError):
+            codec.decode(codec.encode(arr), out)
+
+
+def test_err_out_too_large():
```
jakirkham: Should this be an error?

jakirkham:
So I've added a check to make sure `out` is able to fit the data. If the data does not fit, we still write into `out` (as there doesn't seem to be a way to avoid it), but we raise if we detect more bytes to include. A test is added for the case of a too small `out` as well.

jakirkham:
Is the existing release note sufficient or should we tack this onto that one?

jakirkham:
cc @funkey (for awareness)

numcodecs/gzip.py
```diff
@@ -38,24 +38,35 @@ def encode(self, buf):
...
     def decode(self, buf, out=None):
 
         # normalise inputs
-        buf = ensure_contiguous_ndarray(buf)
-        if out is not None:
-            out = ensure_contiguous_ndarray(out)
+        if PY2:  # pragma: py3 no cover
+            buf = ensure_contiguous_ndarray(buf)
+        else:  # pragma: py2 no cover
+            buf = ensure_bytes(buf)
```
alimanfoo: I'm surprised about this. Is it really necessary to ensure bytes on PY3? That will imply a copy if buf is not bytes. Previously ``ensure_contiguous_ndarray(buf)`` was working on all Pythons?

alimanfoo:
> Is the existing release note sufficient or should we tack this onto that one?

I'd say just adding a reference to this PR in the existing release note would be fine.

numcodecs/gzip.py
jakirkham: Basically it [always copies on Python 2]( https://github.com/python/cpython/blob/2.7/Modules/_io/bytesio.c#L798 ), but will [skip copying on Python 3]( https://github.com/python/cpython/blob/3.7/Modules/_io/bytesio.c#L923 ) if it's already a `bytes` object. So the idea here is if we already have a `bytes` object (e.g. reading from all(?) Zarr stores), then we should try to pass the `bytes` object through unchanged. If we coerce a copy in the process on Python 3, this is no worse as `BytesIO` would have copied it anyways if it wasn't a `bytes` object.

alimanfoo: Ah, got it. Thanks.

alimanfoo: Maybe worth adding a comment here, just so we don't forget why bytes is preferred.

jakirkham: Sure, have added comments to explain why `ensure_contiguous_ndarray` is used in the Python 2 case and why `ensure_bytes` is used in the Python 3 case. HTH

jakirkham:
Have added to the `GZip` release note and included some comments for buffer handling in `GZip.decode`.

alimanfoo:
Great, thanks @jakirkham.

numcodecs/gzip.py
```diff
@@ -38,24 +38,39 @@ def encode(self, buf):
...
+                if decompressor.read(1) != b'':
+                    raise ValueError("Unable to fit data into `out`")
+            else:
+                out = ensure_ndarray(decompressor.read())
 
         # handle destination - Python standard library zlib module does not
         # support direct decompression into buffer, so we have to copy into
         # out if given
-        return ndarray_copy(decompressed, out)
+        return out
```
jakirkham: The comment above should have been dropped. Fixing with PR ( https://github.com/zarr-developers/numcodecs/pull/141 ).

PR #139, opened 2018-11-29
Fix issue with fixture generation

### LINKED ISSUE #138: Error during generation of fixture data
Builds of the numcodecs-feedstock on conda-forge are currently failing for version 0.6.0 (https://github.com/conda-forge/numcodecs-feedstock/pull/28). The problem is within the following test functions:

```
tests/test_astype.py::test_backwards_compatibility FAILED
tests/test_quantize.py::test_backwards_compatibility FAILED
```
The error message is:

```
                # one time encode and save array
                if not os.path.exists(enc_fn):  # pragma: no cover
                    enc = codec.encode(arr)
                    with open(enc_fn, mode='wb') as ef:
>                       ef.write(enc)
E                       ValueError: ndarray is not C-contiguous
```

I think what has happened here is that the fixture data are not being picked up during the conda-forge builds, and so they are being regenerated during the test run. Something has changed since the fixture data was originally generated, and so regeneration of fixture data for the astype and quantize codecs is failing, it looks like on cases where the encoded data returned by the codec is not C-contiguous. Needs a little investigation to determine appropriate fix.

alimanfoo:
Resolves #138 by ensuring that encoded data are converted to bytes before being written to a file.

TODO:

### DIFF
--- a/numcodecs/tests/common.py
+++ b/numcodecs/tests/common.py
@@ -215,6 +215,7 @@ def check_backwards_compatibility(codec_id, arrays, codecs, precision=None, pref
             # one time encode and save array
             if not os.path.exists(enc_fn):  # pragma: no cover
                 enc = codec.encode(arr)
+                enc = ensure_bytes(enc)
                 with open(enc_fn, mode='wb') as ef:
                     ef.write(enc)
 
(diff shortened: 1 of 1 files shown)


numcodecs/tests/common.py
```diff
@@ -215,17 +215,15 @@ def check_backwards_compatibility(codec_id, arrays, codecs, precision=None, pref
...
 
             # load and decode data
             with open(enc_fn, mode='rb') as ef:
                 enc = ef.read()
                 dec = codec.decode(enc)
-                if isinstance(dec, np.ndarray):
-                    dec_arr = dec.reshape(-1, order='A')
-                else:
-                    dec_arr = ensure_ndarray(dec)
+                dec_arr = ensure_ndarray(dec).reshape(-1, order='A')
```
alimanfoo: For the record, this is just a simplification, not needed to address the issue at hand.

jakirkham:
Thanks for tackling this so quickly, @alimanfoo.

PR #143, opened 2018-11-30
Handle (new) buffer protocol conforming types in ``Pickle.decode``

jakirkham:
Python 3's `pickle.loads` is able to handle anything that conforms to the (new) buffer protocol as long as it is C-contiguous. We go ahead and enforce this on all inputs; raising if that is not possible. On Python 2 `cPickle.loads` is not as flexible as it requires a `bytes` object explicitly. This enforces that case as well.

TODO:

### DIFF
--- a/numcodecs/pickles.py
+++ b/numcodecs/pickles.py
@@ -48,9 +48,9 @@ class Pickle(Codec):
         return pickle.dumps(buf, protocol=self.protocol)
 
     def decode(self, buf, out=None):
-        if PY2:
+        if PY2:  # pragma: py3 no cover
             buf = ensure_bytes(buf)
-        else:
+        else:  # pragma: py2 no cover
             buf = ensure_contiguous_ndarray(buf)
 
         dec = pickle.loads(buf)
--- a/numcodecs/pickles.py
+++ b/numcodecs/pickles.py
@@ -6,7 +6,7 @@ import numpy as np
 
 
 from .abc import Codec
-from .compat import PY2, ensure_bytes
+from .compat import PY2, ensure_bytes, ensure_contiguous_ndarray
 
 
 if PY2:  # pragma: py3 no cover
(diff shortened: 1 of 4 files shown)


numcodecs/pickles.py
```diff
@@ -48,6 +48,11 @@ def encode(self, buf):
         return pickle.dumps(buf, protocol=self.protocol)
 
     def decode(self, buf, out=None):
+        if PY2:
+            buf = ensure_bytes(buf)
+        else:
+            buf = ensure_contiguous_ndarray(buf)
```
jakirkham: Just to provide context from the Python codebase, [Python 2 requires `bytes` here]( https://github.com/python/cpython/blob/d4f9cf5545d6d8844e0726552ef2e366f5cc3abd/Modules/_pickle.c#L1179 ) and [Python 3 uses the (new) buffer protocol to get a C-contiguous array]( https://github.com/python/cpython/blob/d4f9cf5545d6d8844e0726552ef2e366f5cc3abd/Modules/_pickle.c#L1179 ).

numcodecs/tests/common.py
```diff
@@ -147,6 +147,10 @@ def check_encode_decode_array(arr, codec):
     codec.decode(enc, out=out)
     assert_array_items_equal(arr, out)
 
+    enc = codec.encode(arr)
+    dec = codec.decode(ensure_ndarray(enc))
+    assert_array_items_equal(arr, dec)
```
jakirkham: Seems this worked for everything except `Pickle` on Python 2 as demonstrated by [this CI failure]( https://travis-ci.org/zarr-developers/zarr/jobs/461902060#L2893 ). This is likely a consequence of the improved buffer handling under the hood, which `MsgPack`, `JSON`, etc. all use already. `Pickle` was the only one not using these functions, which this PR fixes. So testing this generally seems to be fine. All the other codecs already test decoding with `ndarray`.

jakirkham:
Thanks for taking a look.

PR #145, opened 2018-12-01
Upgrade c-blosc to v1.15.0

### LINKED ISSUE #142: Upgrade c-blosc to 1.15.0
Upgrade the c-blosc submodule to version 1.15.0.

alimanfoo:
Upgrade the c-blosc submodule to version 1.15.0. Resolves #142.

TODO:

### DIFF
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -25,7 +25,7 @@ Release notes
   By :user:`John Kirkham <jakirkham>`, :issue:`155`.
 
 * The bundled c-blosc sources have been upgraded to version 1.15.0. By
-  :user:`Alistair Miles <alimanfoo>`, :issue:`142`, :issue:`145`.
+  :user:`Alistair Miles <alimanfoo>` and :user:`John Kirkham <jakirkham>`, :issue:`142`, :issue:`145`.
 
 .. _release_0.6.1:
 
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -24,6 +24,8 @@ Release notes
 * Revert ndarray coercion of encode returned data.
   By :user:`John Kirkham <jakirkham>`, :issue:`155`.
 
+* The bundled c-blosc sources have been upgraded to version 1.15.0. By
+  :user:`Alistair Miles <alimanfoo>`, :issue:`142`, :issue:`145`.
 
 .. _release_0.6.1:
 


alimanfoo:
Looks like this won't work on PY27 on Windows due to https://github.com/Blosc/c-blosc/issues/248. Backing off this for numcodecs 0.6.2.

jakirkham:
Maybe we could try adding our `stdint.h` with [`-FI` (name force include)]( https://docs.microsoft.com/en-us/previous-versions/8c5ztk84(v=vs.140) ). Here’s [an example]( https://github.com/conda-forge/pycrypto-feedstock/blob/master/recipe/meta.yaml#L28 ). Though we can probably use `extra_compile_arg` to pass it in instead of that hack.

alimanfoo:
Good to know that's an option, though I think in this case we can ask to
fix upstream. This has come in to blosclz.c in a recent commit, and all
other source files in c-blosc which include stdint.h do so with
compatibility checks, so I think this probably just slipped past and they'd
want to fix it.

jakirkham:
Sounds good. I sent a PR. So hopefully that helps.

alimanfoo:
Nice one :+1:

jakirkham:
FWIW I'm comfortable either waiting for the new patch release or integrating that patch into our submodule. Would be fine waiting to release Numcodecs 0.6.2 ( https://github.com/zarr-developers/numcodecs/issues/149 ) until this is handled.

alimanfoo:
Cool, let's give it a few days and see if the c-blosc devs have a chance to make a release.

alimanfoo:
I've moved c-blosc forward to incorporate the fix for stdint.h compatibility in blosclz (https://github.com/Blosc/c-blosc/pull/249). If CI passes I think I'll merge, doesn't seem necessary to wait for a release tag upstream.

docs/release.rst
```diff
@@ -24,6 +24,8 @@ Release notes
 * Revert ndarray coercion of encode returned data.
   By :user:`John Kirkham <jakirkham>`, :issue:`155`.
 
+* The bundled c-blosc sources have been upgraded to version 1.15.0. By
+  :user:`Alistair Miles <alimanfoo>`, :issue:`142`, :issue:`145`.
```
alimanfoo:
```suggestion
  :user:`Alistair Miles <alimanfoo>` and :user:`John Kirkham <jakirkham>`, :issue:`142`, :issue:`145`.
```

jakirkham:
SGTM.

IIRC the only other change in there was to their release notes, which we ignore in our builds. So should be fine.

Should we plan on doing the patch release after this lands?

alimanfoo:
> SGTM.
> 
> IIRC the only other change in there was to their release notes, which we ignore in our builds. So should be fine.

Cool, thanks.

> Should we plan on doing the patch release after this lands?

Yes I reckon. I'll make the release now.

jakirkham:
Thanks @alimanfoo!

PR #152, opened 2018-12-03
Use buffer `StringIO` in `GZip.decode` on Python 2

jakirkham:
Extends the strategy from PR ( https://github.com/zarr-developers/numcodecs/pull/150 ) to `GZip`.

In `GZip`'s `decode` method, we knew that `BytesIO` was always copying the data. So we merely ensured the data was in a form that made it easy to copy (i.e. a contiguous buffer). However we have since learned we can do better as the Python 2 implementation of `StringIO` merely takes the buffer without copying it. So use `StringIO` on Python 2 to avoid a copy here. As this support doesn't extend to Python 3, continue using `BytesIO` in that case. Should avoid copies on Python 2 in the general case when using `GZip`'s `decode`; thus, improving performance there.

ref: http://www.hydrogen18.com/blog/unpickling-buffers.html

ref: https://github.com/python/cpython/blob/2.7/Modules/cStringIO.c#L716
ref: https://github.com/python/cpython/blob/2.7/Modules/cStringIO.c#L681
ref: https://github.com/python/cpython/blob/2.7/Modules/cStringIO.c#L160

TODO:

### DIFF
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -15,6 +15,9 @@ Release notes
 * Use (new) buffer protocol in ``MsgPack`` codec `decode()` method.
   By :user:`John Kirkham <jakirkham>`, :issue:`148`.
 
+* Avoid copying into data in ``GZip``'s `decode()` method on Python 2.
+  By :user:`John Kirkham <jakirkham>`, :issue:`152`.
+
 
 .. _release_0.6.1:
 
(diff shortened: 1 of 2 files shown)


jakirkham:
FWIW a full Python 2/3 solution to this exists in PR ( https://github.com/zarr-developers/numcodecs/pull/153 ), but it is incomplete currently. Have placed it there mainly for reference.

For now it would be nice to just get this incremental improvement in.

PR #153, opened 2018-12-03
WIP: Use buffer protocol with `GZip.decode` on Python 2/3

jakirkham:
This is not complete. Am merely putting it up for reference and possibly discussion, but it is not pressing ATM.

Basically creates a custom file object, `MemoryViewIO`, that takes a (new) buffer protocol conforming item and allows users to work with it like an ordinary file object (e.g. `BytesIO`). However it avoids copying the underlying buffer and instead merely views it, which differs from `BytesIO`. This is used with `GzipFile` as the `fileobj` parameter to allow decompression to occur without an additional copy of the full data.

Though there are still some wrinkles to work out. Namely this doesn't work on Python 2 due to some unusual bugs, which are causing test failures. Will probably take some time to get to the bottom of them. The latest commit restricts `MemoryViewIO` usage to Python 3 where it seems to work well. Also test coverage will need to be improved too.

TODO:

### DIFF
--- a/numcodecs/compat.py
+++ b/numcodecs/compat.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 # flake8: noqa
 from __future__ import absolute_import, print_function, division
+import io
 import sys
 import array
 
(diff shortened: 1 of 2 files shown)


numcodecs/compat.py
```diff
@@ -188,3 +189,71 @@ def ndarray_copy(src, dst):
...
+
+    def read(self, size=-1):
+        data = self.peek(size=size)
+
+        self._pos += len(data)
+
+        return data
+
+    def read1(self, n=-1):
+        return self.read1(size=n)
```
alimanfoo:
```suggestion
        return self.read(size=n)
```

jakirkham: Thanks. Fixed.

jakirkham:
After doing some more profiling, the benefits of this strategy are unclear. Especially as calls to `read` must return a `bytes` object (not a `ndarray` object) due to expectations of `GzipFile`, which introduces copies of the original data. Given the complexity that this change is adding and the benefits are not particularly noteworthy, this doesn't really seem worth pursuing further. As such will close it out.

alimanfoo:
Thanks @jakirkham, I'm also happy to punt on this for now..

PR #136, opened 2018-11-29
Return `ndarray`s from `encode` and (where possible) `decode`

jakirkham:
Ensures binary `encode` functions return `ndarray`s generally. This makes the output easier to anticipate and work with. Also allows more unusual outputs like objects' internal buffers to be exposed in a friendly way to end users (while avoiding copying). Where possible also ensure `decode` functions return `ndarray`s.

TODO:
* [x] Unit tests and/or doctests in docstrings
* [x] ``tox -e py37`` passes locally
* [x] ``tox -e py27`` passes locally
* [x] Docstrings and API docs for any new/modified user-facing classes and functions
* [x] Changes documented in docs/release.rst
* [x] ``tox -e docs`` passes locally
* [x] AppVeyor and Travis CI passes
* [x] Test coverage to 100% (Coveralls passes)

PR #155, opened 2018-12-03
Revert ndarray coercion of encode returned data

jakirkham:
This reverts PR ( https://github.com/zarr-developers/numcodecs/pull/136 ) and PR ( https://github.com/zarr-developers/numcodecs/pull/144 ). Based on other discussions ( most notably issue https://github.com/zarr-developers/zarr/issues/348 ), we have decided to revert this to avoid unnecessary copies at the Zarr level particularly when working with in-memory stores.

TODO:

### DIFF
--- a/numcodecs/gzip.py
+++ b/numcodecs/gzip.py
@@ -5,7 +5,7 @@ import io
 
 
 from .abc import Codec
-from .compat import ensure_bytes, ensure_ndarray, ensure_contiguous_ndarray, PY2
+from .compat import ensure_bytes, ensure_contiguous_ndarray, PY2
 
 
 if PY2:  # pragma: py3 no cover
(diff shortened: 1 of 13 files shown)


numcodecs/gzip.py
```diff
@@ -44,11 +44,11 @@ def encode(self, buf):
             compressor.write(buf)
 
         try:
-            compressed = compressed.getbuffer()
+            compressed = ensure_ndarray(compressed.getbuffer())
```
alimanfoo: Do we want to ensure this is bytes here, for consistency with PY3?

jakirkham: The idea here was to avoid a copy by just getting access to the buffer. That said, can understand how having this differ between Python 2/3 might be confusing. If we want to unify them, we should just use `compressed.getvalue()` in both cases.

jakirkham: No worries. Thanks for clarifying. Should be fixed with the latest commit.

jakirkham:
Thanks for the feedback. Should be ready for another look when you have a chance.

alimanfoo:
All good, thanks @jakirkham.

PR #171, opened 2019-02-14
Handle datetime/timedelta case after object case

jakirkham:
As the object check and datetime/timedelta check are exclusive and the object check follows by raising an exception, go ahead and swap the order of these two checks to have the object check go first. That way, we are not spending time running the object check after we have already determined it is datetime or timedelta and casted it to a non-object type array. Instead we merely have verified it is a non-object array if we got to the datetime/timedelta check, which makes it being datetime or timedelta a possibility, but not a certainty.

TODO:


jakirkham:
Went ahead and merged as this was a trivial change. If there are issues, please let me know and I would be more than happy to fix them.

PR #180, opened 2019-03-19
Check cpuinfo only when needed.

QuLogic:
The cpuinfo module does not support s390x and [fails to import](https://kojipkgs.fedoraproject.org//work/tasks/9255/33639255/build.log), but we don't really need it since we've set `CFLAGS`. Move the import later so it's only used when actually checking CPU instruction support.

TODO:
* [N/A] Unit tests and/or doctests in docstrings
* [N/A] Docstrings and API docs for any new/modified user-facing classes and functions

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,6 @@ from __future__ import absolute_import, print_function, division
 from glob import glob
 import os
 from setuptools import setup, Extension
-import cpuinfo
 import sys
 from distutils.command.build_ext import build_ext
 from distutils.errors import CCompilerError, DistutilsExecError, \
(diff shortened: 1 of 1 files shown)


QuLogic:
Since those files require setting `-DSHUFFLE_SSE2_ENABLED` and `-DSHUFFLE_AVX`2_ENABLED`, maybe check if the user's set that in `CFLAGS`?

It's impossible to just look for a compiler option (e.g., `-msse2`), because it could be implied by any number of `-march` flags.

jakirkham:
Would it be possible to raise this with the developers of cpuinfo?

QuLogic:
It's known upstream: https://github.com/workhorsy/py-cpuinfo/issues/69

jakirkham:
It seems like the developer has done some work on it (has a branch) and probably needs more people from the community with access to that architecture to engage with him and try it out. Would you mind engaging in that issue directly and sharing a bit more about what in that branch doesn’t work?

jakirkham:
It looks like upstream addressed this issue. So I've submitted PR ( https://github.com/zarr-developers/numcodecs/pull/202 ), which merely updates `cpuinfo.py` to a commit with the needed feature. Can you please check if that is acceptable @QuLogic? If so, I'll go ahead and merge that PR.

PR #195, opened 2019-07-31
BLD: add newer gcc and clang to build matrix

llllllllll:
Adds the current latest release for gcc and clang

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,15 @@ sudo: false
 
 addons:
   apt:
+    sources:
+      - ubuntu-toolchain-r-test
+      - sourceline: 'deb http://apt.llvm.org/xenial/ llvm-toolchain-xenial-8 main'
+        key_url: 'https://apt.llvm.org/llvm-snapshot.gpg.key'
     packages:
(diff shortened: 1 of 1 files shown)


jakirkham:
This might need to be adjusted to exclude or otherwise handle macOS.

joshmoore:
In light of recent build updates (#320) anything here that needs resurrecting?

jakirkham:
Don't think so.

Travis CI defaulted to fairly old compilers, but we've since moved to GitHub Actions and using conda-forge compilers, which are quite recent and regularly updated.

Also this preceded the addition of wheel builds (in addition to Conda packages, which we already had). So it is a lot easier for users to get binaries without having to build from source.

That said, certainly appreciate the effort here. It was always a bit challenging wrangling with these things on Travis. Hopefully similar issues should be easier to fix now that we manage our own dependencies (including compilers) with Conda.

joshmoore:
Understood & agreed. Thanks from my side as well, @llllllllll. If there's anything that would help you on the GH actions front, please let us know.

PR #212, opened 2019-11-15
WIP: Allow CuPy

jakirkham:
This adds a small change to handle CuPy `ndarrays` in the `compat` functions. Admittedly we may come up with a better way to address the underlying concern here, but I wanted to have this here for tracking. Also it would be useful to have somewhere to discuss this.

TODO:

### DIFF
--- a/numcodecs/compat.py
+++ b/numcodecs/compat.py
@@ -152,6 +152,12 @@ def ensure_bytes(buf):
         if arr.dtype == object:
             raise TypeError('object arrays are not supported')
 
+        # Force CuPy arrays to NumPy arrays
+        # because they don't have a `tobytes` method (yet)
+        # xref: https://github.com/cupy/cupy/pull/2617
+        if isinstance(arr, cp.ndarray):
+            arr = arr.get()
+
         # create bytes
         buf = arr.tobytes(order='A')
 
--- a/numcodecs/compat.py
+++ b/numcodecs/compat.py
@@ -167,6 +167,12 @@ def ensure_bytes(buf):
 def ensure_text(s, encoding='utf-8'):
     if not isinstance(s, text_type):
         s = ensure_contiguous_ndarray(s)
+
+        # Force CuPy arrays to NumPy arrays
+        # as they support the buffer protocol
+        if isinstance(s, cp.ndarray):
+            s = s.get()
+
         s = codecs.decode(s, encoding)
     return s
 
--- a/numcodecs/compat.py
+++ b/numcodecs/compat.py
@@ -6,6 +6,7 @@ import codecs
 import array
 
 
+import cupy as cp
 import numpy as np
 
 
@@ -48,8 +49,8 @@ def ensure_ndarray(buf):
 
     """
 
-    if isinstance(buf, np.ndarray):
-        # already a numpy array
+    if isinstance(buf, (np.ndarray, cp.ndarray)):
+        # already an array
         arr = buf
 
     elif isinstance(buf, array.array) and buf.typecode in 'cu':


numcodecs/compat.py
```diff
@@ -48,8 +49,8 @@ def ensure_ndarray(buf):
 
     """
 
-    if isinstance(buf, np.ndarray):
-        # already a numpy array
+    if isinstance(buf, (np.ndarray, cp.ndarray)):
+        # already an array
         arr = buf
```
jakirkham: This is easy enough to do, but it would be nice if we didn't need to learn about CuPy here.

We will probably need to find another solution for Numcodecs in the near term, but wanted to mention some related things. Namely there is some discussion about generally checking array types in issue ( https://github.com/numpy/numpy/issues/14891 ).

Also mentioned by Matt in the aforementioned issue, Dask has [`is_arraylike`]( https://github.com/dask/dask/blob/31af7f7040643c447a72c87a8f12457094ec15ff/dask/utils.py#L1079-L1101 ), which presents a way of solving this problem. I'm not sure it totally helps us here as we really need to check that there is a data buffer we can grab and use, but it is at least a good starting point.

Another thought I had is we could registered `ndarray` types and their conversion mechanisms with Numcodecs. This could allow users to add their own array types (like CuPy), but not complicate things too much here.

Other thoughts on this would generally be welcome. 🙂

numcodecs/compat.py
```diff
@@ -151,6 +152,12 @@ def ensure_bytes(buf):
         if arr.dtype == object:
             raise TypeError('object arrays are not supported')
 
+        # Force CuPy arrays to NumPy arrays
+        # because they don't have a `tobytes` method (yet)
+        # xref: https://github.com/cupy/cupy/pull/2617
+        if isinstance(arr, cp.ndarray):
+            arr = arr.get()
```
jakirkham: As noted, this is a workaround. The real solution here is PR ( https://github.com/cupy/cupy/pull/2617 ), which should land in CuPy 7.0.0. IOW this can probably be ignored for the sake of this discussion.

numcodecs/compat.py
```diff
@@ -160,6 +167,12 @@ def ensure_bytes(buf):
 def ensure_text(s, encoding='utf-8'):
     if not isinstance(s, text_type):
         s = ensure_contiguous_ndarray(s)
+
+        # Force CuPy arrays to NumPy arrays
+        # as they support the buffer protocol
+        if isinstance(s, cp.ndarray):
+            s = s.get()
```
jakirkham: One way to solve this might be to call `ensure_bytes` here. That would guarantee we have something we can perform text decoding on. It would also generalize for our use case here.

On the flip side that does incur a copy. Though this may be pretty inexpensive in the grand scheme of things.

numcodecs/compat.py
mrocklin: > as we really need to check that there is a data buffer we can grab and use, but it is at least a good starting point.

Perhaps one could check for the presence of a `__array__` or `__cuda_array_interface__` protocol?

numcodecs/compat.py
mrocklin: I think in Dask we do a check like `if hasattr(s, "get"): s = s.get()`.  This is a little non-specific, but it works well in practice and doesn't require the cupy import.

numcodecs/compat.py
mrocklin: Ooh, +1 to `cupy.ndarray..tobytes`

numcodecs/compat.py
jakirkham: That's a good point. Thanks! Will give that a look.

jakirkham:
Yeah the short nature of this change is more a consequence of work Alistair and I did last year (interestingly also in November 🙂). ( https://github.com/zarr-developers/numcodecs/pull/128 )

Thanks for the tips. I'm inclined to agree some small tweaks probably make this easier to integrate.

numcodecs/compat.py
jakirkham: Just for context, this is what the `ensure_bytes` change would look like ( https://github.com/zarr-developers/numcodecs/pull/213 ).

jakirkham: It's also possible we can get away with dropping this part entirely. This function is mainly used in conjunction with reading things (like JSON). So ensuring it can work with CuPy is probably unnecessary.

numcodecs/compat.py
alimanfoo: Yep, if goal here is to ultimately enable codecs with GPU implementations then using ``__cuda_array_interface__`` sounds good. E.g., when using numba cuda functions you get back an instance of some other class (DeviceArray or something like that IIRC) but that exports ``__cuda_array_interface__`` too.

numcodecs/compat.py
alimanfoo: Yes not sure you need to worry about cupy arrays here as this is mainly for JSON reading IIRC.

PR #160, opened 2018-12-21
Developed zfp python wrapper for numcodecs and is ready to be used

halehawk:
Closes #117

Want to know how to add tests. 
Also want to know how to add zfp-0.5.4 c source codes to c-blosc/internal-complibs

TODO:
* [ ] Unit tests and/or doctests in docstrings
* [ ] ``tox -e py37`` passes locally
* [ ] ``tox -e py27`` passes locally
* [ ] Docstrings and API docs for any new/modified user-facing classes and functions
* [ ] Changes documented in docs/release.rst
* [ ] ``tox -e docs`` passes locally
* [ ] AppVeyor and Travis CI passes
* [ ] Test coverage to 100% (Coveralls passes)

PR #229, opened 2020-04-21
numcodecs.zfpy is ready

halehawk:
numcodecs.zfpy is ready with pip install from https://test.pypi.org/simple/ zfpy==0.5.5rc1, this package has no py38 build and osx build.

TODO:
* [x] ``tox -e py38`` passes locally


numcodecs/zfpy.py
```diff
@@ -0,0 +1,90 @@
...
+
+            # do compression
+            comp_arr = _zfpy.compress_numpy(buf, write_header=True, **self.compression_kwargs)
+
+            return comp_arr
+
+        def decode(self, buf, out=None):
+
+            # normalise inputs
+            buf = ensure_bytes(buf)
```
jakirkham, line 68: Why do you need `bytes`? This will copy the data unlike `ensure_contiguous_ndarray` or `ensure_ndarray`.

halehawk:
I tried, but zfpy decompress_numpy API only accept bytes, doesn’t accept numpy.ndarray.

numcodecs/zfpy.py
rabernat: John, can you clarify this. I am confused why the input to `decode` will be an array? The numcodecs [API docs](https://numcodecs.readthedocs.io/en/stable/abc.html) describe `buf` as follows:
> Encoded data. May be any object supporting the new-style buffer protocol or array.array under Python 2.

Should we open an issue in zfpy to inquire whether they could accept an array instead of bytes?

halehawk: I  am asking Peter for a py38 build on linux and windows. It may take a while to get it. In the mean time, can I skip py38 testing and osx testing? I cannot find a way to skip py38 testing for numcodecs.zfpy only. Do you know how to do it? @rabernat @jakirkham

jakirkham: @rabernat @halehawk, `ensure_ndarray` takes a view onto anything that is a new-style buffer and returns it as a NumPy array (also a new-style buffer). Though it is just the same compressed 1-D `uint8` array of data. Mainly it acts as a type checker as it raises on anything that is not a new-style buffer. It also gives us something easier to manipulate. That said, we could cast it to a `memoryview` after if you prefer. The main point is `ensure_ndarray` is basically free whereas `ensure_bytes` always copies (unless it is actually a `bytes` object). Would be better not to copy if we can avoid it both for performance and memory usage reasons.

numcodecs/zfpy.py
```diff
@@ -0,0 +1,90 @@
...
+            if out is not None:
+                out = ensure_contiguous_ndarray(out)
+
+            # do decompression
+            dec = _zfpy.decompress_numpy(buf)
```
jakirkham, line 73: Guessing by the name this requires a NumPy array. Currently we are forcing `buf` to a `bytes` object. Would update the line discussed above to `ensure_ndarray` to fix that.

halehawk:
No, zfpy compress_numpy() returns a bytes object, I used a wrong variable
name here, I will change it.
decompress_numpy has the API as follows:
cpdef np.ndarray decompress_numpy(
bytes compressed_data,
):
So it only accepts bytes object. ensure_bytes(buf) just checks if the buf
is a bytes object or not, and did not copy. So using ensure_bytes(buf), it
works fine on calling compress_numpy and decompress_numpy from my end and
didn't introduce any extra copying.

If I used ensure_contiguous_ndarray(buf), it will convert buf from a bytes
type to a ndarray type, and it cannot be accepted by decompress_numpy API,
and generated error :

TypeError: Argument 'compressed_data' has incorrect type (expected bytes,
got numpy.ndarray)

So I am thinking if I can still use ensure_bytes(buf) here, since it works
well without any performance issue. @*jakirkham*  @rabernat

halehawk:
Good news, zfpy py38 build is available on test.pypi.org now.

halehawk:
numcodecs.zfpy passed all tests on osx ci as well, I didn't skip the tests.

jakirkham:
Do you have a link to their code?

halehawk:
Here is their zfpy.pyx.
https://github.com/LLNL/zfp/blob/release0.5.5/python/zfpy.pyx

halehawk:
Can you approve merging the branch since it passed all tests? @jakirkham @rabernat

rabernat:
It looks like we still have zfp source code in this PR (e.g. zfp.c, zfp.pyx). Can you remove it?

halehawk:
I removed zfp.py zfp.c. @rabernat

numcodecs/zfpy.py
```diff
@@ -0,0 +1,89 @@
...
+        def __repr__(self):
+            if self.mode == _zfpy.mode_fixed_accuracy:
+                r = '%s(mode=%r, tolerance=%s)' % (type(self).__name__, self.mode, self.tolerance)
+            elif self.mode == _zfpy.mode_fixed_rate:
+                r = '%s(mode=%r, rate=%s)' % (type(self).__name__, self.mode, self.rate)
+            elif self.mode == _zfpy.mode_fixed_precision:
+                r = '%s(mode=%r, precision=%s)' % (type(self).__name__, self.mode, self.precision)
+            else:
+                pass
+            return r
```
jhamman, line 89: I think this may need another look. I think the pass should probably just raise an `ValueError`. Or better yet, skip all the if/else checks and return a repr with a longer format:

```python
def __repr__(self):
    r = '%s(mode=%r, tolerance=%s, rate=%s, precision=%s)' % (type(self).__name__, self.mode, self.tolerance, self.rate, self.precision)
    return r
```

You'll note other codecs in this repo do a similar thing: 

https://github.com/zarr-developers/numcodecs/blob/1734bee75f113dde5a59ab38046f9ef158a5c296/numcodecs/lzma.py#L66-L70

halehawk:
I changed zfpy repr to one line. @jakirkham

jakirkham:
> Here is their zfpy.pyx. https://github.com/LLNL/zfp/blob/release0.5.5/python/zfpy.pyx
> […](#)

Makes sense. Copying to `bytes` seems like a reasonable short-term solution. Have raised issue ( https://github.com/LLNL/zfp/issues/96 ) to see if we can avoid this copy in the future.

da-wad:
> Good news, zfpy py38 build is available on test.pypi.org now.

You're welcome :-)

numcodecs/blosc.c
```diff
@@ -304,7 +305,7 @@ END: Cython Metadata */
   #undef BASE
   #undef MASK
   #ifdef SIZEOF_VOID_P
-    enum { __pyx_check_sizeof_voidp = 1 / (int)(SIZEOF_VOID_P == sizeof(void*)) };
+    enum { __pyx_check_sizeof_voidp = 1/(SIZEOF_VOID_P == sizeof(void*)) };
```
jakirkham, line 308: It seems `blosc` is changing here. Can we back out those changes as they are unrelated?

halehawk: This change is same as the current master. Do you want roll back?

jakirkham: Ah didn't realize that. Sorry. Maybe we can merge the branch with `master` to clear this from the diff then?

halehawk:
Which branch? I am waiting for my branch to merge.

jhamman:
@halehawk - I think we need you to merge upstream/master into your branch.

halehawk:
Merged @jhamman @jakirkham

halehawk:
@jakirkham PR #160 closed, please continue your discussion here. @lindstro

halehawk:
Looks like the delocate description here https://stackoverflow.com/questions/47380150/dealing-with-dylibs-when-distributing-python-packages can solve your problem. @lindstro @jakirkham

lindstro:
@halehawk That's the same SO post I referenced.  delocate already is part of the [osx multibuild script](https://github.com/matthew-brett/multibuild/blob/devel/osx_utils.sh) used to build zfpy wheels on Travis, but from the [Travis logs](https://travis-ci.org/github/LLNL/zfpy-wheels/jobs/692349615) I cannot tell if it's actually being executed.

halehawk:
@lindstro, I see your manylinux called build_index_wheel then called
repair_wheelhouse, but your osx didn't call repair_wheelhouse anywhere.

docs/release.rst
```diff
@@ -5,7 +5,9 @@ Release notes
 
 0.7.0
 -----
-
+* The :class:`numcodecs.zfpy.ZFPY` codec is now supported on Python 3.8 if
```
joshmoore, line 8: This will need moving to 0.7.2 (or later)

jakirkham (approved):
LGTM

Tidied a few minor things. Hope that is ok 🙂

Thoughts from others (@jhamman @rabernat)?

numcodecs/zfpy.py
rabernat: This conversation does not seem resolved. Have we decided to keep `ensure_bytes`? Have we discussed this choice with the zfp devs?

numcodecs/zfpy.py
rabernat: This comment does not seem resolved to me.

halehawk:
No development on zfp devs on this issue yet.

halehawk:
*@rabernat * currently, we don't have a lot of cases to use zfpy data with
storage required ensure_contiguous_ndarray. Can we have this code as a
temporal solution and update it in the future?

rabernat (approved):
I'm happy to see this merged as is. However, I would like someone to open a new issue to help us track the specific question of `ensure_bytes` vs `ensure_ndarray`. Eventually we want to get to the point where zfpy can pass a new-style buffer directly to numcodecs.

rabernat:
Should we merge this PR?

halehawk:
I really want to, but Peter has not released a new zfp release yet which
has the ensure_ndarray feature. I am just wondering if we can create a
conda package which can include this feature in a minor version update.

rabernat:
I don't think the `ensure_ndarray` issue is a dealbreaker for this coded. I think we should just get it out there and then implement this in a future release. This effort has been going for several years!

halehawk:
Sounds good to me.

halehawk:
What do you think? @jakirkham @jhamman

jakirkham:
Yep I think I already approved this PR

halehawk:
Do we need one more approval? Or can one of you merge it?

requirements_dev.txt
```diff
@@ -1,3 +1,5 @@
 Cython==0.29.21
 msgpack==1.0.2
 numpy==1.19.0
+zfpy==0.5.5
```
jakirkham, line 4: Could we restrict this to Python 3.8 or older?

requirements_rtfd.txt
```diff
@@ -6,3 +6,4 @@ numpydoc
 mock
 numpy
 cython
+zfpy==0.5.5
```
jakirkham, line 9: Same as above

jakirkham (commented):
CI appears to be failing on Python 3.9. Guess it is just due to lack of wheels on that platform. Suggested some changes above to fix that

lindstro:
> CI appears to be failing on Python 3.9. Guess it is just due to lack of wheels on that platform. Suggested some changes above to fix that

I can add 3.9 wheels.  Someone recently requested this, but at the time I wasn't sure if we require MB_ML_VER=2010 on Linux as we do for 3.8.  I can't seem to find the thread where this was discussed.  Does anyone know?

jakirkham:
Honestly I think a new release of zfpy  is probably more important than Python 3.9 support. I don’t know how much work that would be though

halehawk:
I used  zfpy==0.5.5; python_version <'3.9'
Looks like it worked, zfpy was ignored.

But I got error at Build Docs
autodoc: failed to import class 'ZFPY' from module 'numcodecs.zfpy'; the
following exception was raised:
93Traceback (most recent call last):
94 File
"/opt/hostedtoolcache/Python/3.9.2/x64/lib/python3.9/site-packages/sphinx/util/inspect.py",
line 393, in safe_getattr
95 return getattr(obj, name, *defargs)
96
AttributeError: module 'numcodecs.zfpy' has no attribute 'ZFPY'

How can I skip it?

lindstro:
> Honestly I think a new release of zfpy is probably more important than Python 3.9 support. I don’t know how much work that would be though

We've essentially frozen new features and are working on tests and documentation for zfp 0.5.6.  Shouldn't be too much longer.  But once we do the release, it would make sense to include Python 3.9.

jakirkham:
Yep that makes sense. Knew that work towards a release was ongoing. So didn't want to ask for more things, but yeah agree that it makes sense to support Python 3.9 at the same time

halehawk:
@jakirkham, I tried autodoc_mock_imports, it doesn't work. Can I remove
zfpy.rst temporarily?

jakirkham:
Could we just skip building the docs on Python 3.9?

halehawk:
I cannot figure out how to skip building the docs on Python 3.9 only, I
added zfpy to

MOCK_MODULES = ['msgpack','zfpy']
in conf.py.
At least now, all tests passed.

jakirkham:
Thank you @halehawk! 😄

pjpetersik:
@rabernat and @halehawk What is the reason that the ZFPY Codec actually uses `ensure_contiguous_ndarray` instead of `ensure_ndarray`?

When I was playing around with this Codec for compression of weather model data, i.e. 2m temperature from ICON-EU, replacing `ensure_contiguous_ndarray` with `ensure_ndarray` did not led to any errors and led to higher compression ratios.

halehawk:
@pjpetersik, do you mean ensure_contiguous_ndarray in zfpy.encode?
@pjpetersik if the zfp.compress_numpy which is called in zfpy.encode gets a
higher dimension of array, it can get a better compression ratio. From your
description, it looks like you got this situation.

@jakirkham is ensure_contiguous_ndarray supposed to keep the original shape
of the array? Or does it convert the original array to a flat array?

pjpetersik:
@halehawk I meant `ensure_contiguous_ndarray` in the `ZFPY` Codec class:
https://github.com/zarr-developers/numcodecs/blob/e364ceb8daef1f9a30e50a0aa1f44c7c983e08a7/numcodecs/zfpy.py#L58

jakirkham:
Many of the compressors we use don't know how to handle shape correctly hence why it does that

We could add a flag to `ensure_contiguous_ndarray` to skip the `reshape` step. Then use that flag here

halehawk:
@jakirkham Will you add the flag? Then I can use the flag in the encode.
Thanks!

jakirkham:
@halehawk it probably will be a while before I have time to do that. In the midst of doing several release atm. If you or @pjpetersik , would like to do a PR though would be happy to review :)

halehawk:
I can try to do a PR, just hope all the build tests can be easily
passed. @jakirkham @pjpetersik

halehawk:
@jakirkham I tried to add a flag, then I failed the test on f-contiguous
array. Can I flat f-contiguous arrays but not c-contiguous arrays?

jakirkham:
Maybe it would be best to move this discussion to a new issue where it is easier to track

PR #232, opened 2020-05-20
Use pytest to parametrize some of the tests.

Carreau:
This basically move the looping on multiple conditions from being the
test responsibility to being pytest's it has several advantages.

1) I find it make the test more redable as you only have to focus on
   what you want to test.

2) In case of test failures it will have more informative message with
   which sub-condition fails and will not stop on the first error,
   usually allowing to narrow down the error case faster.

3) It allow to run a single subtest when fixing code (using the --lf, or
   --ff pytest flags).

4) Show which subcase are slow with --durations=..., and could integrade
   with multi-worker testing or randomisations.

It does have minimal drawbacks; but this does increase some of the setup
time which is now done sometime for each case instead of one per group.

use_threads seem to be used a lot, so I made it a module-scoped fixture,
a function only have to use the parameter use_threads to automatically
been generic over some parameters, other are ad-hoc.

I did not parametrize on everything as I mostly want to test if you are
ok with this approach.

This was initially push by trying to test current decode_partial in
blosc which is still in progress and was failing on a subset of cases.

--- 

GitHub note, the diff will me messy if you don't "ignore whitespace", most of the changes are de-indent and removal of for loops.

TODO:

### DIFF
--- a/numcodecs/tests/test_blosc.py
+++ b/numcodecs/tests/test_blosc.py
@@ -1,4 +1,3 @@
-import itertools
 from multiprocessing import Pool
 from multiprocessing.pool import ThreadPool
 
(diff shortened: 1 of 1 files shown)


Carreau:
I'm slightly confised the test show as still pending but are actually passing.

Carreau:
closing and reopening to kick the tests.

Carreau:
rebased.

PR #249, opened 2020-09-14
Clean fixtures

avalentino:
The PR adds some cleanup code to ``check_backwards_compatibility`` to remove test artefacts form the source tree.

TODO:

### DIFF
--- a/numcodecs/tests/common.py
+++ b/numcodecs/tests/common.py
@@ -302,6 +302,10 @@ def check_backwards_compatibility(codec_id, arrays, codecs, precision=None, pref
                     assert_array_equal(arr, dec_arr)
                     assert arr_bytes == ensure_bytes(dec)
 
+    if os.path.exists(fixture_dir):
+        import shutil
+        shutil.rmtree(fixture_dir)
+
 
 def check_err_decode_object_buffer(compressor):
     # cannot decode directly into object array, leads to segfaults


jakirkham:
As we actually commit the fixture files to track compatibility, I don't think we want to delete them as part of the tests. That said, it does appear we are missing the `Base64` fixtures, which PR ( https://github.com/zarr-developers/numcodecs/pull/251 ) fixes. Please let us know if there is something else we are overlooking here.

avalentino:
@jakirkham thank you for the quick reply.
Sorry, I did not understand that the behavior was intended, my fault.
My problem is that build artefacts create problems when I generate the debian package for numvodecs.
Anyway I think I can solve the issue with some small change to the build script so closing this PR and sorry for the noise.

joshmoore:
@avalentino : is there anything else that needs to be kept in sync / done on a numcodecs release to simplif the debian packaging?

avalentino:
Hi, @joshmoore packaging of numcodecs is already quite straightforward.
Currently there are 4 patches applied (see https://salsa.debian.org/science-team/numcodecs/-/tree/master/debian/patches) but I'm going to remove the last one ASAP in consideration of what discussed in this issue.

Probably the only think that could be improved upstream form a packaging point of view is to provide also the possibility to built numcodecs using system version of the blosc library (instead of using the bundled sources).
This is currently addressed by https://salsa.debian.org/science-team/numcodecs/-/blob/master/debian/patches/0003-No-embedded-libs.patch.

jakirkham:
Thanks for the additional context, Antonio. Also thanks for working on the Debian package.

If there are issues around including these binaries in the package, happy to discuss options as well. I think we mainly went with committing them to the repo as it was easier to get started and maintain. Though we can revisit if there are other needs.

Yeah I understand that need to use external libraries as well. This is something I've been wanting to do for conda-forge too. If you have a patch already that is working for this, would be happy to review a PR here.

avalentino:
@jakirkham is short in Debian of course we do not want to include test artefacts in the package, and also we need to leave the source tree clean after the build. Both the requirements can be controlled by the package makefile as explained in a previous comment so non need to change anything on your side.
I opened the PR because I incorrectly thought the behaviour was not intended.

Regarding the patch for external libraries, as I said, in Debian we use this one:  https://salsa.debian.org/science-team/numcodecs/-/blob/master/debian/patches/0003-No-embedded-libs.patch.
Please note that it is a quick patch that could also work for Conda but IMHO it is at least incomplete in the general case.
The missing part is a mechanism that allows the user to control the behaviour.
I suppose you want to keep using bundled blocs sources by default and eventually allow the user to use a command line option or an environment variable to change the behaviour and use system libs only In specific cases.
the part regarding the command line option or the environment variable is totally missing in the linked patch.

PR #265, opened 2021-01-06
Try to migrate to Gh-action.

Carreau:
As we are out of time on .travis-ci for the zarr org.

[Description of PR]

TODO:

### DIFF
--- a/.github/workflows/ci-linux.yaml
+++ b/.github/workflows/ci-linux.yaml
@@ -1,0 +1,41 @@
+name: Linux CI
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: linux
+    strategy:
+      fail-fast: false
+      matrix:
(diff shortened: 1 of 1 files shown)


jakirkham:
Thanks for starting this Matthias! 😄

It looks like GitHub Actions doesn't provide us a compiler atm. Not sure how we add one. Maybe a simple solution would be to use the same image we use for wheels here. That way we are building in the same environment we later use for deployment. Though open to other ideas if you have thoughts 🙂

Carreau:
I've tried to fix osx; but was unable to without using conda. See #266 for part of a solution.
I'm not sure if building wheels on conda is fine or not and would lead to issues on non-conda installs.

jakirkham:
Are there changes from here that we would like to integrate as well?

joshmoore:
There doesn't seem to be anything here that's not in https://github.com/zarr-developers/numcodecs/blob/main/.github/workflows/ci-linux.yaml

PR #266, opened 2021-01-06
Try to fix osx ci-build.

Carreau:
Not sure why it was not working anymore, migrate to conda.


Carreau:
As seen in #265 OSX ci now fails on github action; I was not able to fix it without conda, and newer compilers. I'm not sure how to make it work with 3.5. Is it time to drop 3.5 for numcodecs ?

jakirkham:
Yeah dropping Python 3.5 is fine. We already did so in Zarr ( https://github.com/zarr-developers/zarr-python/pull/603 ). It's also past EOL.

Carreau:
I tried to update bloc to see if that fix the build issue, but then it fail a it can't find snappy-c.h which has been remove post 0.18.1 I'm not sure what we should do (sent env to disable snappy or how to re-add the source ??

Carreau:
I have no clue, I give up for today; one way would be to remove `Werror` which I don't even know where it's coming from

jakirkham:
Ok maybe we should hold off on updating Blosc then. I may have missed, but what was motivating the Blosc update? Was there another issue prompting it or were we just wanting to try updating?

Carreau:
> I may have missed, but what was motivating the Blosc update? Was there another issue prompting it or were we just wanting to try updating?

Trying to see if that fixes the error in osx-wheels; I'll revert all changes I've made to try to fix the wheel building, and maybe we can take care of osx wheels in a separate PR.

jakirkham:
FWIW Jacob has also been doing some work with GitHub Actions and figured out how to interactively debug issues. Maybe that would help us figure out what is happening on CI? Here's a link to the blogpost he wrote up about this.

https://jacobtomlinson.dev/posts/2021/how-to-interactively-debug-github-actions-with-netcat/

Carreau:
Thanks for the link, I would have also try to setup reverse ssh but that is easier. But right now it fails even on my machine for different reasons so I'll need to fix that first maybe ? 

For now I've reverted the changes to the wheel building so that at least we have a partially working CI.

I'm going to deactivate the travis-ci requirement as well

Carreau:
All test that are expecting to pass are passing with the exception of: 
  - coverall (won't be triggered as travisis not working anymore)
  - OSX wheel; which is in a separate PR. I'll go ahead on merge and iterate on other PRs to try to fix things.

PR #267, opened 2021-01-13
Try to fix osx wheel build

Carreau:
On top of #266 only look at fixing wheel build on OS for now.


Carreau:
OSX Wheel is working, only remaining issue is coverage that won't be fix until we migrate away from travis. 

Merging.

jakirkham:
Thanks Matthias! 😄

Issue #260, opened 2020-11-26
Add support for byte shuffle as a codec so that it can be used as a Zarr filter


pbranson:
**Enhancement Request**

Following on from discussion here:
https://github.com/intake/fsspec-reference-maker/issues/11 
It would be great to expose the blosc library shuffle operations as a numcodec Codec so that shuffle could be included as a zarr filter. 

This will assist with efforts to expand the functionality provided by the fsspec ReferenceFileSystem to a broader range of datasets stored in hdf format on S3. 

The plan would be to expose https://github.com/Blosc/c-blosc/blob/9fae1c9acb659159321aca69aefcdbce663e2374/blosc/shuffle.h as a cython shuffle.pyx module in numcodecs.

If this sounds like a good enhancement I would be happy to try making a PR for this.

alimanfoo:
Hi @pbranson, just to say this sounds like a good idea to me, PR welcome.

rsignell-usgs:
@pbranson , this is exciting!  I think many NetCDF4 files "in the wild" will have shuffle on as that appears to be the default for the [NCO tools](https://sourceforge.net/p/nco/discussion/9829/thread/5de98550/#549f).

pbranson:
Thanks for the encouragement @rsignell-usgs ! Have made a start but getting hit by a pre-christmas crunch. Hopefully have a PR ready for review by early next week.

pbranson:
I have had a go at exporting the internal blosc shuffle functions, however the buffers used by those functions use an unsigned char* pointer, presumably due to the byte shuffle, this presents some warnings during compile.

 
The c function header is: 
`blosc_internal_shuffle(const size_t bytesoftype, const size_t blocksize,
                       const uint8_t* _src, const uint8_t* _dest)`

I have some branches of numcodecs 
https://github.com/pbranson/numcodecs/tree/shufflecodec 
and blosc-c
https://github.com/pbranson/c-blosc/tree/exportshuffle
which compile, however the encode just returns a byte array filled with 0's using the following test code:

```
import numpy as np
from numcodecs import blosc
from numcodecs.blosc import Blosc, Shuffle

codec = Shuffle()
arr = np.random.normal(loc=1000, scale=1, size=(10,))
enc = codec.encode(arr)
dec = codec.decode(enc)

print(arr)
print(enc)
print(dec)
```
which outputs:
```
[ 999.11031606 1001.00931215  998.04921042  998.75484377  999.72977095
[...]

PR #270, opened 2021-01-24
Build + Support Py39 and Various CI Updates

JacksonMaxfield:
I was originally trying to get `numcodecs` for py39 + macOS building [as I have a build failing](https://github.com/AllenCellModeling/aicsimageio/pull/185/checks?check_run_id=1755420390), but, in working on this:

* I saw https://github.com/zarr-developers/numcodecs/pull/265 (Updating CI / moving from Travis to GitHub Actions)
* I also saw https://github.com/zarr-developers/numcodecs/issues/269 (blosc upgrade)
  * But then saw in zarr dev community call notes that it may be preferred to move from blosc submodule to blosc pip install.
  * And all the tests + wheel builds on MacOS + Py39 so not sure where to fall on this

In addressing the CI changes, I tried to figure out the `travis` and `appveyor` prior setups but it was a bit tricky. The one missing thing is `coveralls` which has a [GitHub Action](https://github.com/marketplace/actions/coveralls-github-action) that we could use but didn't know how you wanted to handle that. (Sidenote: other option is to move to codecov which is _very_ easy to setup in my experince)

Reference Issues:
https://github.com/zarr-developers/numcodecs/issues/262
https://github.com/zarr-developers/numcodecs/issues/261
https://github.com/zarr-developers/numcodecs/issues/269

TODO:
* [x] Unit tests and/or doctests in docstrings
* [x] ``tox -e py39`` passes locally
* [x] Docstrings and API docs for any new/modified user-facing classes and functions
* [x] Changes documented in docs/release.rst
* [x] ``tox -e docs`` passes locally
* [ ] Test coverage to 100% (Coveralls passes)

PR #273, opened 2021-02-17
Add support for byte shuffle as a codec so that it can be used as a Zarr filter

pbranson:
PR for a byte shuffle codec as discussed in https://github.com/zarr-developers/numcodecs/issues/260

Became too difficult (for me atleast) to see a way forward making use of the optimised blosc-c shuffle. 

This add a dependency on numba which may not be acceptable, in which case perhaps a cython implementation of the shuffle functions will be required. Would appreciate advice if that is the case

TODO:

- [x] `tox -e py39` passes locally
- [x] `tox -e docs` passes locally


joshmoore:
@pbranson : The wheel GH actions don't make use of requirements.txt. I think you'll need to register `numba` in `setup.py` as well.

pbranson:
Thanks for the suggestion @joshmoore. The wheels now build for all except Python 3.9

joshmoore:
Interesting:
```
ModuleNotFoundError: No module named 'numpy'
```
Looks like NumPy is missing somewhere, but I haven't tracked down where yet.

joshmoore:
At least for Windows/3.9, it looks like it's failing on trying to _build_ the numba wheel. If there's no wheel available, perhaps installing numba via conda would help? But this raises the issue of whether end users will face the same issue.

pbranson:
Looks like the build problems with 3.9 as numba does not yet have a release with 3.9 support (see https://github.com/numba/numba/issues/6345). This is coming soon in numba 0.53.0

However, it would probably be good to get some consensus on whether including numba as a dependency is something acceptable to for numcodecs: @jakirkham, @alimanfoo do you have any guidance on that front?

martindurant:
I would suggest, to push this forward, that numba **not** be a dependency, but only needed when instantiating this filter, i.e., using any other filters works fine without numba installed. However, numba should be in the tests, and installed from conda. There is a good reason for the existence of conda: you can include complex binary dependencies like LLVM outside of the python tree in a way that is really hard to achieve with pip.

How hard is it to write shuffle cython, like the other filters?

numcodecs/shuffle.py
```diff
@@ -0,0 +1,109 @@
...
+# Copyright by The HDF Group.                                                #
+# All rights reserved.                                                       #
+#                                                                            #
+# This file is part of HSDS (HDF5 Scalable Data Service), Libraries and      #
+# Utilities.  The full HSDS copyright notice, including                      #
+# terms governing use, modification, and redistribution, is contained in     #
+# the file COPYING, which can be found at the root of the source code        #
+# distribution tree.  If you do not have access to this file, you may        #
+# request a copy from help@hdfgroup.org.                                     #
+##############################################################################
```
jakirkham, line 14: Do we want to be lifting their code and including it here? Is there a way to implement this that does not rely on their code? Seems like this raises some licensing concerns

numcodecs/shuffle.py
```diff
@@ -0,0 +1,109 @@
...
+@jit(nopython=True)
+def _doUnshuffle(src, des, element_size):
+    count = len(src) // element_size
+    for i in range(element_size):
+        offset = i*count
+        e = src[offset:(offset+count)]
+        for byte_index in range(count):
+            j = byte_index*element_size + i
+            des[j] = e[byte_index]
+    return des
```
jakirkham, line 43: Does NumPy have some mechanism for doing this directly?

setup.py
```diff
@@ -311,7 +311,7 @@ def run_setup(with_extensions):
             'setuptools-scm>1.5.4'
         ],
         install_requires=[
-            'numpy>=1.7',
+            'numpy>=1.7', 'numba==0.53.0rc2'
```
jakirkham, line 314: Agree with Martin. It would be good to soften this requirement or drop it completely.

Another thing to consider would just be using an optional 3rd party dependency for the algorithm as a whole like ( https://github.com/kiyo-masui/bitshuffle ), which would allow users to get a performant implementation that we can leverage indirectly

jakirkham (commented):
Thank for looking into this Paul! 😄

Unfortunately have a few concerns with this implementation noted below. There are probably a few different ways to address this. I think the easiest will likely be adding a new optional dependency for this functionality. Though maybe other solution like using NumPy directly may be viable (though I haven't dug deeply into this)

martindurant:
I made a cython version in https://github.com/pbranson/numcodecs/pull/1 that eliminated the dep problem. It's not quite done, but the tests pass. Almost identical code. 

For licensing, presumably we need to just find a publicly free definition of shuffle - I'm sure HDF didn't invent it. Actually I'm not sure what their license is anyway.

setup.py
pbranson: I had also found this library but didnt know if it would be compatible for _byte_ shuffling, which is what is what appears to be most often applied with netCDF.

numcodecs/shuffle.py
pbranson: Probably using some indexing operations, however I expect it would be much slower, but haven't benchmarked

setup.py
pbranson: @martindurant previously suggested raising with the cblosc devs if they would be willing to make their shuffle functions available through a public interface - perhaps that is the best long term approach?

numcodecs/shuffle.py
pbranson: Yep for sure, I just wanted to be transparent. Perhaps @jreadey could offer some guidance. I think that the shuffle algorithm would be considered public domain? In which case can we just rename some variables ;-)

pbranson:
Thanks @martindurant and @jakirkham for taking a look at this!

The license for HSDS is apache 2.0 https://github.com/HDFGroup/hsds/blob/master/LICENSE which from my naive reading seems to suggest that inclusion of the copyright notice is all that is needed...

@martindurant Thanks a lot for the cython code! I have had a look at your PR - will have a go at merging it into this branch to resolve the numba dep

pbranson:
Yay, all ticks!

numcodecs/shuffle.pyx
```diff
@@ -0,0 +1,99 @@
...
+# cython: language_level=3
+
+import numpy as np
+from .compat import ensure_bytes, ensure_contiguous_ndarray, ndarray_copy, ensure_ndarray
+from .abc import Codec
+
+
+cdef _doShuffle(const unsigned char[:] src, unsigned char[:] des, int element_size):
+    cdef int count, i, j, offset, byte_index
+    count = len(src) // element_size
```
martindurant, line 14: I did a cythonize-annotate, and *this* line has extra code because element_size could be zero and raise an exception; except that we know that it cannot the way we have called it.

The rest I have optimized to
```cython
cimport cython

@cython.boundscheck(False)
@cython.wraparound(False)
cdef void _doShuffle(const unsigned char[::1] src, unsigned char[::1] des, int element_size) nogil:
    cdef int count, i, j, offset, byte_index
    count = len(src) // element_size
    for i in range(count):
        offset = i*element_size
        for byte_index in range(element_size):
            j = byte_index*count + i
            des[j] = src[offset + byte_index]


@cython.boundscheck(False)
@cython.wraparound(False)
cdef void _doUnshuffle(const unsigned char[::1] src, unsigned char[::1] des, int element_size) nogil:
    cdef int count, i, j, offset, byte_index
    count = len(src) // element_size
    for i in range(element_size):
        offset = i*count
        for byte_index in range(count):
            j = byte_index*element_size + i
            des[j] = src[offset+byte_index]
```

Tests pass and are much faster. Should we have a case where we know what the output should be, rather than just checking round-trips?

martindurant (commented):
Sorry, my review comment went long...

martindurant:
Aha, can add `@cython.cdivision(True)` to the two c functions to avoid the possibility of making a python `ZeroDivisionError`

martindurant:
https://stackoverflow.com/a/19576057/3821154 suggests that the cdivision thing may not be worthwhile; in the functions, we only do the division the once. I guess it should be measured.

martindurant:
Closes https://github.com/intake/fsspec-reference-maker/issues/1

jakirkham:
Would change all the `int` indices, lengths, etc. to `Py_ssize_t`. That's what Python uses for shapes and the like

jakirkham:
Enabling `cdivision` makes sense. We don't need the error checking AFAICT. If we do we should handle that in the Python end. Would also mark any C functions as `nogil` to avoid calling overhead

martindurant:
> Would also mark any C functions as nogil to avoid calling overhead

Yes, I suggested this, although actually it *adds* a very little overhead in the case that we are sure this is to be called single-threaded. I would suggest that, with these changes, we are ready to go. The only thing extra would be a test where we have known outputs, rather than just testing the round-trip. I can add the code changes in a PR to @pbranson 's branch, but I don't actually know how to make a test case, as I don't know what the algorithm really does!

jakirkham:
Yep just agreeing with you 🙂 The small perf change is not too surprising

Occasionally there are people using the threaded Dask Scheduler. That said we would have to use the `with nogil` block to actually release the GIL. Marking the function `nogil` just removes some boilerplate in C and allows the function to be called in `with nogil` blocks. IOW `nogil` functions are more similar to `nopython` in Numba (so `nogil` is maybe a misnomer here)

Agreed tests and fixtures would be helpful. Yeah I think this is something we need Paul's help on 🙂

numcodecs/__init__.py
```diff
@@ -92,6 +92,13 @@
 from numcodecs.base64 import Base64
 register_codec(Base64)
 
+try:
+    from numcodecs import shuffle
+    from numcodecs.shuffle import Shuffle
```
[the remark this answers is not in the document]
pbranson: yeah I thought the same, not sure what is going on but the tests fail on all platforms except linux if I dont include the exception handling (e.g. https://github.com/zarr-developers/numcodecs/pull/273/checks?check_run_id=2091910819)

numcodecs/shuffle.pyx
```diff
@@ -0,0 +1,100 @@
...
+
+import numpy as np
+from .compat import ensure_contiguous_ndarray
+from .abc import Codec
+
+cimport cython
+
+@cython.boundscheck(False)
+@cython.wraparound(False)
+cdef void _doShuffle(const unsigned char[::1] src, unsigned char[::1] des, int element_size) nogil:
```
jakirkham, line 15:
```suggestion
cdef void _doShuffle(const unsigned char[::1] src, unsigned char[::1] des, Py_ssize_t element_size) nogil:
```

numcodecs/shuffle.pyx
```diff
@@ -0,0 +1,100 @@
...
+    for i in range(count):
+        offset = i*element_size
+        for byte_index in range(element_size):
+            j = byte_index*count + i
+            des[j] = src[offset + byte_index]
+
+
+@cython.boundscheck(False)
+@cython.wraparound(False)
+cdef void _doUnshuffle(const unsigned char[::1] src, unsigned char[::1] des, int element_size) nogil:
```
jakirkham, line 27:
```suggestion
cdef void _doUnshuffle(const unsigned char[::1] src, unsigned char[::1] des, Py_ssize_t element_size) nogil:
```

jakirkham (commented):
`element_size` is likely small. That said, everything else is `Py_ssize_t`. So this should save us from casting

numcodecs/__init__.py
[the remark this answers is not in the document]
pbranson: @jakirkham do you have any advice on this?

numcodecs/shuffle.pyx
pbranson: Thanks Martin, incorporated with typing suggestions from John

numcodecs/shuffle.pyx
```diff
@@ -0,0 +1,100 @@
...
+        out.view(buf.dtype)[:len(buf)] = buf[:]
+        return out  # no shuffling needed
+
+    buf_size = buf.nbytes
+    if buf_size % element_size != 0:
+        raise ValueError("Shuffle buffer is not an integer multiple of elementsize")
+
+    _doUnshuffle(buf.view("uint8"), out.view("uint8"), element_size)
+
+    return out
```
jakirkham, line 62: I wonder if we can move this into the `Shuffle` codec. Would simplify the code here a bit and avoid the overhead of these intermediate functions

pbranson: Yep good point. refactored accrodingly

numcodecs/shuffle.pyx
```diff
@@ -0,0 +1,94 @@
...
+
+    """
+
+    codec_id = 'shuffle'
+
+    def __init__(self, elementsize=4):
+        self.elementsize = elementsize
+
+    def encode(self, buf, out=None):
+        buf = ensure_contiguous_ndarray(buf)
```
martindurant, line 53: Most of `encode` and `decode` is the same, could make a separate method like `_prepare_array`. This would shorten the code and, I suspect, have no impact on performance. Only a mild suggestion.

numcodecs/shuffle.pyx
```diff
@@ -0,0 +1,94 @@
...
+    cdef Py_ssize_t count, i, j, offset, byte_index
+    count = len(src) // element_size
+    for i in range(element_size):
+        offset = i*count
+        for byte_index in range(count):
+            j = byte_index*element_size + i
+            des[j] = src[offset+byte_index]
+
+
+class Shuffle(Codec):
```
jakirkham, line 37: Would separate the `Shuffle` class into a Python file (say `shuffle.py`) and then leave the Cython bits in `_shuffle.pyx` (underscore as it is internal). This should avoid Cythonizing a class we don't really need Cythonized (and thus avoid any potentially odd behavior that may surface from that). Though would need to mark the function above as `cpdef` to ensure they expose a Python callable API

numcodecs/shuffle.py
pbranson: I think this can be resolved, the cython code is mostly rewritten compared to using numba

numcodecs/__init__.py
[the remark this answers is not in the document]
pbranson: Refactored to separate cython code from python class

numcodecs/shuffle.pyx
pbranson: Thanks, have added _prepare_array suggestion. The flow control seems a little odd with the return, possibly a more elegant way to handle the input checking...

numcodecs/shuffle.pyx
pbranson: Thanks, have separated out the cython and python as suggested, but on platforms other than linux still getting import failures due to the cython shuffle functions not being recognised. See https://github.com/zarr-developers/numcodecs/pull/273/checks?check_run_id=2113066075 

Not sure how to resolve this other than to wrap the import in the try...catch during the __init__.py

numcodecs/__init__.py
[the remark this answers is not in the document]
pbranson: This still has errors on platforms other than linux....

martindurant:
Does that imply that it doesn't import/work on those platforms, or just that the first import is odd?

pbranson:
Heh, I went into the logs for the Windows CI and for all of the codecs that have that ImportError handling the tests are skipped...
https://github.com/zarr-developers/numcodecs/runs/2092033530#step:6:12 

So on Windows only 108 out of a possible 593 tests are actually executed.... So I dont know if it works on those platforms! Doesnt seem satisfactory

martindurant:
cc ^ @zarr-developers/core-devs

joshmoore:
I assume related to https://github.com/zarr-developers/numcodecs/pull/270

martindurant:
cc @JacksonMaxfield - it seems we get `ImportError`s for any cython implementation on windows, and plenty of tests not running. Is this expected?

JacksonMaxfield:
The windows builds were definitely testing nearly everything on my PR, including Cython based deps / sections of the code.

I know some of this is already said above but condensing into a single comment as I get caught up:

See [here](https://github.com/zarr-developers/numcodecs/pull/270/checks?check_run_id=1764712461) any of the Windows CI builds.

> ... collected 108 items / 6 skipped / 102 selected
> ... 107 passed, 6 skipped, 1 xfailed, 4 warnings in 7.60s

Did 400 tests get added to this repo in that past month? Did I miss something in my PR?

No that's not the case because even on that same PR the ubuntu and mac OS builds have the expected number of tests detected and ran.

> ... collected 533 items
> ... 494 passed, 39 xfailed, 4 warnings in 11.61s

Well then whats the difference between the windows build from the others?

Taking the step definition of the ci-macos and the ci-windows workflows to a text compare shows they are exactly the same....

So no differences in the CI workflow definition, and crucially it uses the same testing command as the others, so still confused as to what is going on.

There is this comment from my PR: https://github.com/zarr-developers/numcodecs/pull/270#issuecomment-767010371

So, if my memory is correct and understanding the context of that comment is correct, there was _no_ Windows testing prior to my PR.

A bit of digging on "pytest fails to pick up Cython tests on Windows" got me to this [stackoverflow](https://stackoverflow.com/questions/49033350/how-to-use-cython-with-pytest)

Which has a final update of:
> How can I use pytest with Cython? How can I discover why these methods aren't working and then fix it?
>
> FINAL UPDATE: After taking both the original problem and the question Updates into consideration (thanks @hoefling for solving these issues!), this question is now reduced to the question of:
> 
>> why can pytest no import the Cython module calculateScore, even though running the code just with python (no pytest) works just fine?

Looks like this has been an issue of pytest + cython + windows for a while?

JacksonMaxfield:
One more update. Looked at that stackover again and saw there was a single upvoted answer down below that linked to this answer: https://stackoverflow.com/a/49068163

Which says to build the cython extension manually instead of using `cythonize` but that seems cumbersome?

martindurant:
The question is: can we be certain that these cythonized modules will work on windows? We know from the linux tests that the algorithms themselves are fine.

hoefling:
I'm not sure why I was summoned by Github, but I took a look at the workfow YAMLs, and this is my guess for the different amount of tests on Win + MacOS on one side, and Linux on the other:

Linux: project is installed properly via `pip install --editable .`, Win + Mac: only `setup.py build` is executed. This leaves the extensions in the `build` dir, so they can't be imported. Whereas the editable install runs `setup.py build_ext --inplace` under the hood, which copies built extensions to the target package.

Second, Linux uses `pytest -v --cov=numcodecs --doctest-modules --doctest-glob=*.pyx numcodecs` which explicitly runs doctests, while Win + Mac uses `pytest -v --pyargs numcodecs`, so no doctests (they are not being run by default). why using `pyargs` at all here, BTW?

Hope this helps!

JacksonMaxfield:
Hey @hoefling! Sorry for pinging you. You and the person who helped answer the question on StackOverflow have the same name (or you _are_ the same person!) and when I copy-pasted the answer in to GitHub I forgot to remove the `@` character but your solutions definitely work for running all of the tests on Windows and Mac!

Made a PR to fix this: https://github.com/zarr-developers/numcodecs/pull/276

pbranson:
Thats awesome, thanks everyone!

For this PR should I change the codec registration to not handle the import error and we accept that the windows CI tests will fail until #276 is merged?

hoefling:
> Hey @hoefling! Sorry for pinging you. You and the person who helped answer the question on StackOverflow have the same name (or you _are_ the same person!) and when I copy-pasted the answer in to GitHub I forgot to remove the `@` character but your solutions definitely work for running all of the tests on Windows and Mac!

Ah, I get it now, funny thing :-) Absolutely no worries, happy to help!

martindurant:
I have merged #276 , so you can merge in, and then we can merge this too (note the conflict in release.rst).

pbranson:
OK I think this is ready!

docs/index.rst
```diff
@@ -96,6 +97,8 @@ documentation, code reviews, comments and/or ideas:
 * :user:`Trevor Manz <manzt>`
 * :user:`Grzegorz Bokota <Czaki>`
 * :user:`Josh Moore <joshmoore>`
+* :user:`Martin Durant <mdurant>`
```
martindurant, line 100: If this is github username, I am @martindurant

docs/release.rst
```diff
@@ -6,6 +6,11 @@ Release notes
 Unreleased
 ----------
 
+* Add support for Python 3.9 and Update GitHub Actions.
+
+* Add support for byte Shuffle filter
+  By :user:`Paul Branson <pbranson>` and :user:`Martin Durant <mdurant>` :issue:`273`.
```
martindurant, line 12: Same again

martindurant:
Is it right that all the fixture .dat and .npy files show up as empty?

docs/release.rst
pbranson: Oops sorry about that. Fixed

pbranson:
Yeah that is weird about the fixtures - they arent empty on my system - if you click to view file it shows as ~7kb raw. Actually now they are all showing correctly?!

martindurant:
Screen Shot 2021-03-18 at 11 44 42

But if I go to view raw and download it, it has content. I suppose the tests woudn't pass without them!

jakirkham:
Yeah I think one has to manually add them. They shouldn't be empty, but we do have them in `.gitignore` I believe. @pbranson could you please open a new PR that adds these?

martindurant:
@jakirkham - the files really are there, they just didn't show in the diff for some reason
https://github.com/zarr-developers/numcodecs/blob/master/fixture/shuffle/array.00.npy

jakirkham:
Oh interesting did not realize that. Thanks for correcting me Martin 😄

pbranson:
Yay, thanks everyone for your help and time reviewing. I'm pretty new to this so has been a good learning experience!

Issue #269, opened 2021-01-22
Can't install numcodecs on OSX Big Sur


d70-t:
#### Minimal, reproducible code sample, a copy-pastable example if possible

```bash
pip install numcodecs
```

#### Problem description

Installation fails due to Blosc/python-blosc#229. Updating the `c-blosc` submodule to v1.21.0 fixes the issue.

#### Version and installation information

Please provide the following:
* Value of ``numcodecs.__version__``: current master branch
* Version of Python interpreter: 3.9
* Operating system (Linux/Windows/Mac): Mac

jakirkham:
Thanks for the data point! 😄

adamjstewart:
I'm also seeing this problem on macOS Catalina. The issue is with Apple Clang 12.0.0, not with the specific macOS version. See https://github.com/Blosc/c-blosc/issues/316 for my error msg.

kindjacket:
I'm getting the same issue on a M1 MacBook Pro MacOS Big Sur 11.5 when trying to install numcodec v0.8.0

The error message is 
```
c-blosc/internal-complibs/zlib-1.2.8/gzread.c:30:15: error: implicit declaration of function 'read' is invalid in C99
```

@d70-t - how did you force the numcodecs install to use a newer version of c-blosc?

kindjacket:
The issue is slightly different when installing in a Python 3.9.6 virtualenv.  Instead I get
```
 c-blosc/internal-complibs/zlib-1.2.8/gzlib.c:256:24: error: implicit declaration of function 'lseek' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
              state->start = LSEEK(state->fd, 0, SEEK_CUR);
                             ^
      c-blosc/internal-complibs/zlib-1.2.8/gzlib.c:14:17: note: expanded from macro 'LSEEK'
      #  define LSEEK lseek
                      ^
      c-blosc/internal-complibs/zlib-1.2.8/gzlib.c:256:24: note: did you mean 'fseek'?
      c-blosc/internal-complibs/zlib-1.2.8/gzlib.c:14:17: note: expanded from macro 'LSEEK'
      #  define LSEEK lseek
                      ^
      /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/stdio.h:162:6: note: 'fseek' declared here
      int      fseek(FILE *, long, int);
               ^
[...]

PR #274, opened 2021-02-18
WIP: pure python

joshmoore:
Following on from https://github.com/zarr-developers/numcodecs/pull/259#issuecomment-759702459, I dropped the blosc module and am slowly making changes in the background to use native blosc wheels, turning this into a pure Python module. If anyone would like to add commits, please feel free (**especially** if you know what you are doing unlike myself...)

TODO:

- [ ] Unit tests and/or doctests in docstrings
- [ ] `tox -e py39` passes locally
- [ ] Docstrings and API docs for any new/modified user-facing classes and functions
- [ ] Changes documented in docs/release.rst
- [ ] `tox -e docs` passes locally
- [ ] GitHub Actions CI passes
- [ ] Test coverage to 100% (Coveralls passes)

PR #259, opened 2020-11-24
Bump c-blosc to 1.20.1

joshmoore:
Attempt bumping c-blosc to fix OSX builds. See https://github.com/zarr-developers/numcodecs/pull/258#issuecomment-731708306 from @avalentino 

* [ ] AppVeyor and Travis CI passes
* [ ] Test coverage to 100% (Coveralls passes)

PR #283, opened 2021-07-28
c-blosc upgrade 1.18.1 -> 1.21.0

kindjacket:
c-blosc submodule upgrade 1.18.1 -> 1.21.0 to fix issue #269 

TODO:


kindjacket:
When trying to run on my local machine I got the following error
```
      c-blosc/blosc/blosc.c:31:12: fatal error: 'snappy-c.h' file not found
        #include "snappy-c.h"
                 ^~~~~~~~~~~~ `
```

kindjacket:
This fixes the file not found snappy issue when compiling c-blosc https://github.com/zarr-developers/numcodecs/pull/283/commits/6df653b85399d107975ace938b7c1b25c16f82d5

IDK - how this will play with other machines.

setup.py
```diff
@@ -79,7 +79,7 @@ def blosc_extension():
     include_dirs += [d for d in glob('c-blosc/internal-complibs/*/*/*')
                      if os.path.isdir(d)]
     define_macros += [('HAVE_LZ4', 1),
-                      ('HAVE_SNAPPY', 1),
+                      # ('HAVE_SNAPPY', 1),
```
jakirkham, line 82: Yeah I don't think we want to skip using Snappy. Did this move somewhere else? Is there a path we need to update?

jakirkham: Hmm...looks like they dropped Snappy

https://github.com/Blosc/c-blosc/pull/295

jakirkham:
```suggestion
                      ('HAVE_SNAPPY', 0),  # Blosc 1.19.0+ dropped Snappy
```

kindjacket: I'll make this change and see if installation still succeeds

kindjacket: Confirming that it did not work with the above change https://github.com/zarr-developers/numcodecs/pull/283#discussion_r679439206 . I get the same error https://github.com/zarr-developers/numcodecs/pull/283#issuecomment-888787956

jakirkham: Just to confirm, that's with `0` and not `1`, correct? If so, that's really weird

kindjacket: yep set to 0. Yeah I can try None too.

kindjacket: Yep same issue with None. I can dig into how the macro passes information to the c-blosc build process

kindjacket: Oh it seems like @joshmoore started on this very same PR but stopped because of a call https://github.com/zarr-developers/numcodecs/pull/259 

Do we want to proceed with this change or go down another path?

joshmoore: @kindjacket: explaining the #259 / #274 decisions -- we discussed that on that call that spending our time on the pure python implementation was probably a bigger win. However, as you can see in my #274, that turned out to be more work than (I) expected.

At this point, I'm inclined to say we temporarily drop snappy in order to get M1 fixed and then re-evaluate once that's working with the options being continue on the #274 path, or to first work on getting snappy working with c-block 1.21+.

@zarr-developers/core-devs : any objections?

joshmoore:
@kindjacket : looks like we have a :thumbsup: from @jakirkham as well (https://github.com/zarr-developers/numcodecs/issues/268), any thoughts on getting this ready to merge?

joshmoore:
_whew_ Starting to look greenish.

jakirkham (approved):

jakirkham:
Actually can we leave the fixture data (instead of deleting it)? We will want this when we add back snappy (somehow 🤔)

joshmoore:
> Actually can we leave the fixture data (instead of deleting it)? We will want this when we add back snappy

Good point. I didn't realize at first that it was an ordered list etc. etc. so I thought deletion was the appropriate way forward. I'll resurrect.

joshmoore:
Any thoughts on version number with the drop of snappy? `0.8.0` I assume.

joshmoore:
I'm taking that at least as no objections.

jakirkham:
Thanks Josh! 😄 

Sorry had to drop 😴 Though happy to trust your judgement on versions 🙂

Raised issue ( https://github.com/zarr-developers/numcodecs/issues/285 ) about readding Snappy

joshmoore:
np.

A heads up that while testing https://github.com/zarr-developers/zarr-python/pull/773 I'm pretty sure that this PR changed the checksum of chunks in the zarr_implementations test.

Issue #288, opened 2021-10-05
Add support to release Linux aarch64 wheels


odidev:
Problem
---------
On aarch64, ‘pip install numcodecs’ builds the wheels from source code and then installs it. It requires the user to have a development environment installed on his system. Also, it takes some time to build the wheels than downloading and extracting the wheels from pypi.

Resolution
-----------
On aarch64, ‘pip install numcodecs’ should download the wheels from pypi 

@joshmoore and Team Please let me know your interest in releasing aarch64 wheels. I can help in this.

odidev:
I have done some modifications to build the wheel for arm64 but were getting below error -
```
gcc -pthread -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -fPIC -DHAVE_LZ4=1 -DHAVE_ZLIB=1 -DHAVE_ZSTD=1 -Ic-blosc/blosc -Ic-blosc/internal-complibs/zlib-1.2.8 -Ic-blosc/internal-complibs/zstd-1.4.8 -Ic-blosc/internal-complibs/lz4-1.9.3 -Ic-blosc/internal-complibs/zstd-1.4.8/dictBuilder -Ic-blosc/internal-complibs/zstd-1.4.8/common -Ic-blosc/internal-complibs/zstd-1.4.8/decompress -Ic-blosc/internal-complibs/zstd-1.4.8/legacy -Ic-blosc/internal-complibs/zstd-1.4.8/dll -Ic-blosc/internal-complibs/zstd-1.4.8/deprecated -Ic-blosc/internal-complibs/zstd-1.4.8/compress -Ic-blosc/internal-complibs/zstd-1.4.8/dll/example -I/opt/_internal/cpython-3.6.12/include/python3.6m -c numcodecs/blosc.c -o build/temp.linux-aarch64-3.6/numcodecs/blosc.o -mno-sse2 -mno-avx2
    gcc: error: unrecognized command line option ‘-mno-sse2’
    gcc: error: unrecognized command line option ‘-mno-avx2’
    [numcodecs] command 'gcc' failed with exit status 1
    Traceback (most recent call last):
      File "/opt/_internal/cpython-3.6.12/lib/python3.6/distutils/unixccompiler.py", line 118, in _compile
        extra_postargs)
      File "/opt/_internal/cpython-3.6.12/lib/python3.6/distutils/ccompiler.py", line 909, in spawn
        spawn(cmd, dry_run=self.dry_run)
[...]

PR #289, opened 2021-10-06
Add linux aarch64 wheel build support

odidev:
Add linux aarch64 wheel build support. 

Related to https://github.com/zarr-developers/numcodecs/issues/288 @joshmoore Could you please review this PR?

### DIFF
--- a/.github/workflows/wheel.yaml
+++ b/.github/workflows/wheel.yaml
@@ -4,13 +4,18 @@ on: [push, pull_request]
 
 jobs:
   build_wheels:
-    name: Build wheel on ${{ matrix.os }}
+    name: Build wheel on ${{matrix.arch}} for ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
+        arch: [auto]
+        include:
+          - os: ubuntu-latest
+            arch: aarch64
     env:
+      CIBW_ARCHS_LINUX: ${{matrix.arch}}
       CIBW_TEST_COMMAND: pytest --pyargs numcodecs
       CIBW_TEST_REQUIRES: pytest
       CIBW_SKIP: "*27* pp* *35*"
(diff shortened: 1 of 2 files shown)


jthetzel:
What is needed to include this pull request in the master branch? Aarch64 wheels for Linux are convenient.

joshmoore:
Apologies, @odidev. I have no idea how I missed this originally! Never hesitate to re-ping or jump on https://gitter.im/zarr-developers/community -- I'll take a look ASAP.

.github/workflows/wheel.yaml
```diff
@@ -27,11 +32,23 @@ jobs:
...
+        name: Set up QEMU
+
       - name: Install cibuildwheel
         run: |
           python -m pip install cibuildwheel==1.8.0 wheel
 
+      - name: Build wheel for aarch64
+        if: ${{ matrix.arch == 'aarch64' }}
+        env:
+          CIBW_ENVIRONMENT: 'DISABLE_NUMCODECS_AVX2=1 DISABLE_NUMCODECS_SSE2=1'
```
joshmoore, line 46: Could these have also be set in the setup.py rather than `pass`ing? Just wondering if having a single place to look would prevent future issues.

joshmoore (commented):
One small comment. Otherwise, seems like we should give it a try. @jakirkham?

joshmoore:
Also looking at the conflict, @odidev and @jthetzel, any thoughts on the use of pypa/cibuildwheel@v2.3.1 from #309?

jthetzel:
> Also looking at the conflict, @odidev and @jthetzel, any thoughts on the use of pypa/cibuildwheel@v2.3.1 from #309?

Testing that at https://github.com/zarr-developers/numcodecs/pull/315 . Please advise if not kosher. I get confused on best practices pull requesting against pull requests.

joshmoore:
@jthetzel: no worries. Git can take care of all of that. More important is the semantics of the explicit calls to cibuildwheel versus the pre-built action which I'm not familiar enough with to deciper ("without more effort"™️)

jakirkham:
Sorry for the lack of reply here. Thought I had already replied.

The main concern from my perspective is CI time is getting pretty long for the wheel builds ( https://github.com/zarr-developers/numcodecs/issues/312 ). Would like to limit the builds for custom architectures, libc implementations, etc.. Namely not running on every commit, but only running if a label is added, some text in the commit, particular files changed, etc. Something to keep the build time manageable.

rabernat:
@jakirkham - have we decided not to support this wheel? Or have we decided we need specific changes to our build system in order to support it?

Basically looking for a recommendation on the way forward.

jakirkham:
Going to close in favor of PR ( https://github.com/zarr-developers/numcodecs/pull/315 ), which is more up-to-date. Hopefully this more cleanly organizes discussion around this change. Thanks for the work @odidev! 🙏

PR #299, opened 2021-12-17
Bitround Codec

### LINKED ISSUE #298: Support new "bitinformation" codec in numcodecs
There is an exciting new paper by @milankl in Nature Computational Science entitled [Compressing atmospheric data into its real information content](https://www.nature.com/articles/s43588-021-00156-2)

### Abstract

> Hundreds of petabytes are produced annually at weather and climate forecast centers worldwide. Compression is essential to reduce storage and to facilitate data sharing. Current techniques do not distinguish the real from the false information in data, leaving the level of meaningful precision unassessed. Here we define the bitwise real information content from information theory for the Copernicus Atmospheric Monitoring Service (CAMS). Most variables contain fewer than 7 bits of real information per value and are highly compressible due to spatio-temporal correlation. Rounding bits without real information to zero facilitates lossless compression algorithms and encodes the uncertainty within the data itself. All CAMS data are 17× compressed relative to 64-bit floats, while preserving 99% of real information. Combined with four-dimensional compression, factors beyond 60× are achieved. A data compression Turing test is proposed to optimize compressibility while minimizing information loss for the end use of weather and climate forecast data.

### More relevant description of the method

> Based on the bitwise real information content, we suggest a strategy for the data compression of climate variables. First, we diagnose the real information for each bit position. Afterwards, we round bits with no significant real information to zero, before applying lossless data compression. This allows us to minimize information loss but maximize the efficiency of the compression algorithms.
> 
> Bits with no or only little real information (but high entropy) are discarded via binary round-to-nearest as defined in the IEEE-754 standard. This rounding mode is bias-free and therefore will ensure global conservation of the quantities that are important in climate model data. Rounding removes the incompressible false information and therefore increases compressibility. Although rounding is irreversible for the bits with false information, the bits with real information remain unchanged and are bitwise reproducible after decompression. Both the real information analysis and the rounding mode are deterministic, also satisfying reproducibility.
>
> Lossless compression algorithms can be applied efficiently to rounded floating-point arrays (the round + lossless method). Many general-purpose lossless compression algorithms are available and are based on dictionaries and other statistical techniques to remove redundancies. Most algorithms operate on bitstreams and exploit the correlation of data in a single dimension only, so we describe such methods as one-dimensional (1D) compression. Here, we use the Zstandard algorithm for lossless compression, which has emerged as a widely available default in recent years.
>
> ...
> In an operational setting we recommend the following workflow. First, for each variable, the bitwise real information content is analyzed from a representative subset of the data. For example, a single time step can be representative of subsequent time steps if the statistics of the data distribution are not expected to change. From the bitwise real information, the number of mantissa bits to preserve 99% of information is determined (the ‘keepbits’). Second, during the simulation, the arrays that will be archived are rounded to the number of keepbits (which are held fixed) and compressed. The first step should be done offline—once in advance of a data-producing simulation. Only the second step has to be performed online, meaning every time a chunk of data is archived.

## Prior Conversation

The compressor is currently implemented in Julia. I emailed Milan to ask about possible paths for supporting this compression codec in numcodecs. I am now moving the discussion here so other devs can weigh in.

I posed the question in the following way

> We would like to investigate implementing this compressor in Numcodecs we can can use it with Zarr data. In order to do that, we would need to either
> - call Julia directly from numcodecs (certainly possible but would limit usage)
> - reimplement the algorithm in python (not appealing)
> - Export a c-compatible binary library from Julia (possible but perhaps difficult) 
> 

Milan replied

> Thanks so much for your interest! I’d be more than happy to develop [BitInformation.jl](https://github.com/milankl/BitInformation.jl) more towards a reference implementation for the analysis of the bitwise information content. But you are quite right that maintaining similar code in several languages might be tricky and unnecessary additional effort. Before making a suggestion to either of these options let me highlight that there are technically two parts to our “compression method”. Quotation marks as the idea of the real bitwise information is certainly also that it should be possible to combine it with other existing compression methods, as we do with Zfp in our paper.
> 
> part 1. The bitwise information analysis. This just informs how many bits should be retained for a given array, and does not have to be repeated, say for following time steps
>
> part 2. The removal of false information. We simply round this to zero with round-to-nearest, while this doesn’t technically compress it can be combined with any lossless algorithm and I see you have plenty in numcodecs already.
>
> So I think we’d need two things in numcodecs, for 
>
> part 1 the bitinformation function from BitInformation.jl  If there’s a way to pull that one directly from Julia into numcodecs that’d be great because then I can keep updating / maintaining the BitInformation.jl and you’d get automatically the newest version.
>
> part 2 a binary round to nearest. I see that np.round rounds in decimal, but we’d need a binary rounding mode to guarantee that the trailing bits are zero. This is technically [3 lines of code](https://github.com/milankl/BitInformation.jl/blob/b335571f5d56c12df579c39ccca1bc29b681f61e/src/round.jl#L19-L21) and so that can be done maybe directly within C/Cython. 
>
> So I’d suggest to reimplement part 2 (the rounding) as it’s only a small piece of code, but we should check how to call Julia efficiently for part 1.

Since we already support ZFP in numcodecs, I would say that this strategy makes a lot of sense

I'm copying Fig. 1 from the paper into this issue, since it provides a great visualization of the method

image

## Next Steps

I would note that there is also a final step in this method, which is to apply a lossless compression codec to 

As a concrete step forward, we could imagine first implementing the "binary rounding" as a standalone codec in numcodecs.

One technical question I have for @milankl is the following: The [round](https://github.com/milankl/BitInformation.jl/blob/b335571f5d56c12df579c39ccca1bc29b681f61e/src/round.jl#L15-L22) function has a signature (translated to python like)

```python
def round(x: float, setmask: int, shift: int, shavemask: int) -> float:
    ...
```

where `x` is the array to be rounded. _Can you confirm that `setmask`, `shift`, and `shavemask` are **constants** for each variable, which are determined based on the mutual information analysis from "part 1"?_ If this is the case, we can simply treat these as scalar parameters for the bitwise rounding codec.

This also suggests an even simpler path to supporting this compressor in python:
- we do the mutual information analysis in Julia using BitInformation.jl (only needs to be done once for each array)
- export the scalar parameters, e.g. via copy / paste or a simple text file
- create the appropriate codec chain in numcodecs, e.g. `bitwise_rounding` + `zstd` or `zst`

I'm very excited about this opportunity and really eager to try it out, e.g. on the [LLC4320 ocean simulations](https://medium.com/pangeo/petabytes-of-ocean-data-part-1-nasa-ecco-data-portal-81e3c5e077be).

rabernat:
Eventually closes #298 

TODO:

### DIFF
--- a/numcodecs/bitround.py
+++ b/numcodecs/bitround.py
@@ -12,6 +12,9 @@ class BitRound(Codec):
         self.keepbits = keepbits
 
     def encode(self, buf):
+        # TODO: figure out if we need to make a copy
+        # Currently this appears to be overwriting the input buffer
+        # Is that the right behavior?
         a = ensure_ndarray(buf).view()
         assert a.dtype == np.float32
         b = a.view(dtype=np.int32)
--- a/numcodecs/bitround.py
+++ b/numcodecs/bitround.py
@@ -1,0 +1,28 @@
+import numpy as np
+
+
+from .abc import Codec
+from .compat import ensure_ndarray, ndarray_copy
+
+
+class BitRound(Codec):
+    codec_id = 'bitround'
+
+    def __init__(self, keepbits: int):
+        self.keepbits = keepbits
+
+    def encode(self, buf):
+        a = ensure_ndarray(buf).view()
+        assert a.dtype == np.float32
+        b = a.view(dtype=np.int32)
+        maskbits = 23 - self.keepbits
+        mask = (0xFFFFFFFF >> maskbits) << maskbits
+        half_quantum1 = (1 << (maskbits - 1)) - 1
+        b += ((b >> maskbits) & 1) + half_quantum1
+        b &= mask
+        return b
+
+    def decode(self, buf, out=None):
+        data = ensure_ndarray(buf).view(np.float32)
+        out = ndarray_copy(data, out)
+        return out
--- a/numcodecs/tests/test_bitround.py
+++ b/numcodecs/tests/test_bitround.py
@@ -3,18 +3,76 @@ import numpy as np
 import pytest
 
 from numcodecs.bitround import BitRound
-from numcodecs.tests.common import check_encode_decode, check_config, \
-    check_repr, check_backwards_compatibility
 
 # adapted from https://github.com/milankl/BitInformation.jl/blob/main/test/round_nearest.jl
 
 
+# TODO: add other dtypes
 @pytest.fixture(params=[np.float32])
 def dtype(request):
     return request.param
 
 
-def test_smoke(dtype):
-    data = np.zeros((3, 2), dtype=dtype)
-    codec = BitRound(keepbits=10)
-    codec.decode(codec.encode(data))
+# number of mantissa bits for each dtype
+MBITS = {np.float32: 23}
+
+
+def round(data, keepbits):
+    codec = BitRound(keepbits=keepbits)
+    data = data.copy()  # otherwise overwrites the input
+    encoded = codec.encode(data)
+    return codec.decode(encoded)
+
+
+def test_round_zero_to_zero(dtype):
+    a = np.zeros((3, 2), dtype=dtype)
+    # Don't understand Milan's original test:
+    # How is it possible to have negative keepbits?
+    # for k in range(-5, 50):
+    for k in range(0, MBITS[dtype]):
+        ar = round(a, k)
+        np.testing.assert_equal(a, ar)
+
+
+def test_round_one_to_one(dtype):
+    a = np.ones((3, 2), dtype=dtype)
+    for k in range(0, MBITS[dtype]):
+        ar = round(a, k)
+        np.testing.assert_equal(a, ar)
+
+
+def test_round_minus_one_to_minus_one(dtype):
+    a = -np.ones((3, 2), dtype=dtype)
+    for k in range(0, MBITS[dtype]):
+        ar = round(a, k)
+        np.testing.assert_equal(a, ar)
+
+
+# This triggers a 'negative shift count' error in the codec
+def test_no_rounding(dtype):
+    a = np.random.random_sample((300, 200)).astype(dtype)
+    keepbits = MBITS[dtype]
+    ar = round(a, keepbits)
+    np.testing.assert_equal(a, ar)
+
+
+APPROX_KEEPBITS = {np.float32: 10}
+
+
+# This does not pass at the default tolerance of allclose
+# How is it different from Julia's ≈ operator?
+def test_approx_equal(dtype):
+    a = np.random.random_sample((300, 200)).astype(dtype)
+    ar = round(a, APPROX_KEEPBITS[dtype])
+    np.testing.assert_allclose(a, ar)
+
+
+def test_idempotence(dtype):
+    a = np.random.random_sample((300, 200)).astype(dtype)
+    for k in range(20):
+        ar = round(a, k)
+        ar2 = round(a, k)
+        np.testing.assert_equal(ar, ar2)
+
+
+# TODO: implement tie_to_even and round_to_nearest
--- a/numcodecs/tests/test_bitround.py
+++ b/numcodecs/tests/test_bitround.py
@@ -1,0 +1,20 @@
+import numpy as np
+
+import pytest
+
+from numcodecs.bitround import BitRound
+from numcodecs.tests.common import check_encode_decode, check_config, \
+    check_repr, check_backwards_compatibility
+
+# adapted from https://github.com/milankl/BitInformation.jl/blob/main/test/round_nearest.jl
+
+
+@pytest.fixture(params=[np.float32])
+def dtype(request):
+    return request.param
+
+
+def test_smoke(dtype):
+    data = np.zeros((3, 2), dtype=dtype)
+    codec = BitRound(keepbits=10)
+    codec.decode(codec.encode(data))


numcodecs/bitround.py
```diff
@@ -0,0 +1,31 @@
...
+    def encode(self, buf):
+        # TODO: figure out if we need to make a copy
+        # Currently this appears to be overwriting the input buffer
+        # Is that the right behavior?
+        a = ensure_ndarray(buf).view()
+        assert a.dtype == np.float32
+        b = a.view(dtype=np.int32)
+        maskbits = 23 - self.keepbits
+        mask = (0xFFFFFFFF >> maskbits) << maskbits
+        half_quantum1 = (1 << (maskbits - 1)) - 1
```
rabernat, line 23: This section will error with `ValueError: negative shift count` if `keepbits=23`. That causes the `test_no_rounding` test to error.

numcodecs/tests/test_bitround.py
```diff
@@ -0,0 +1,78 @@
...
+
+APPROX_KEEPBITS = {np.float32: 10}
+
+
+# This does not pass at the default tolerance of allclose
+# How is it different from Julia's ≈ operator?
+def test_approx_equal(dtype):
+    a = np.random.random_sample((300, 200)).astype(dtype)
+    ar = round(a, APPROX_KEEPBITS[dtype])
+    np.testing.assert_allclose(a, ar)
```
rabernat, line 67: Trying to understand why this test [works in Julia](https://github.com/milankl/BitInformation.jl/blob/ba587d9565864139af18e9ae68b189e98586514f/test/round_nearest.jl#L71-L82) but not as implemented here.

rabernat: I think I found the answer - https://docs.julialang.org/en/v1/base/math/#Base.isapprox

> For real or complex floating-point values, if an atol > 0 is not specified, rtol defaults to the square root of eps of the type of x or y, whichever is bigger (least precise). This corresponds to requiring equality of about half of the significand digits. Otherwise, e.g. for integer arguments or if an atol > 0 is supplied, rtol defaults to zero.

numcodecs/tests/test_bitround.py
```diff
@@ -0,0 +1,81 @@
...
+
+def test_approx_equal(dtype):
+    a = np.random.random_sample((300, 200)).astype(dtype)
+    ar = round(a, APPROX_KEEPBITS[dtype])
+    # Mimic julia behavior - https://docs.julialang.org/en/v1/base/math/#Base.isapprox
+    rtol = np.sqrt(np.finfo(np.float32).eps)
+    # This gets us much closer but still failing for ~6% of the array
+    # It does pass if we add 1 to keepbits (11 instead of 10)
+    # Is there an off-by-one issue here?
+    np.testing.assert_allclose(a, ar, rtol=rtol)
```
rabernat, line 70: The fact that this test passes if we use keepbits=11 instead of 10 makes me think we are dealing with a off-by-one issue, perhaps related to julia vs. python indexing

numcodecs/bitround.py
rkouznetsov: This is a feature. half_quantum1 is indeed inexpressible if your quantum is the least significant bit... I would just add return from the beginning or not call the sub at all if keepbits == 23.  If  keepbits > 23 (or  < 0) you might wish to raise an exception... 
I wonder if the test is needed at all...

milankl: I find a condition like `keepbits >= 0` reasonable, but note that `keepbits < 0` within reasonable limits (see below) shouldn't cause problems as this will just round the exponent bits. E.g.
```julia
julia> using BitInformation
julia> round(4f0,-1)
2.0f0
```
So in this format only ...,0.125,0.5,2,8,... are representable. However, as the exponent bits describe logarithmically distributed floats, round to nearest is now round to nearest in log-space. Meaning that while 4 is round down to 2 in the example above,
```julia
julia> round(4.1f0,-1)
8.0f0
```
although in lin-space 4.1 is closer to 2 than to 8. Sure, this has to be treated with caution, as NaN/Inf are defined through their exponents, meaning you end up with situations like
```julia
julia> round(NaN32,-1)
-0.0f0
julia> round(Inf32,-1)
-0.0f0
```
And the carry bit can propagate into the sign bit (for |x|>2), which is also weird
```julia
julia> round(4f0,-8)
-0.0f0
julia> round(-4f0,-8)
0.0f0
```

numcodecs/tests/test_bitround.py
milankl: To be honest, I just guessed the tolerances here. The motivation for this test was less to test exactness, but just to flag immediately if rounding is completely off.

numcodecs/bitround.py
```diff
@@ -0,0 +1,31 @@
...
+class BitRound(Codec):
+    codec_id = 'bitround'
+
+    def __init__(self, keepbits: int):
+        self.keepbits = keepbits
+
+    def encode(self, buf):
+        # TODO: figure out if we need to make a copy
+        # Currently this appears to be overwriting the input buffer
+        # Is that the right behavior?
```
milankl, line 17: In BitInformation.jl the rounding is implemented as scalar version which does not overwrite the input (as float32 is immutable so a copy is created anyway), however, I define a rounding function for arrays, that can either act in-place (i.e. overwriting the bits in an existing array) or acts on a copy of the array, such that the input array in unchanged.

numcodecs/tests/test_bitround.py
```diff
@@ -0,0 +1,81 @@
...
+    codec = BitRound(keepbits=keepbits)
+    data = data.copy()  # otherwise overwrites the input
+    encoded = codec.encode(data)
+    return codec.decode(encoded)
+
+
+def test_round_zero_to_zero(dtype):
+    a = np.zeros((3, 2), dtype=dtype)
+    # Don't understand Milan's original test:
+    # How is it possible to have negative keepbits?
```
milankl, line 30: You just end up rounding the exponent bits, see other comment

numcodecs/tests/test_bitround.py
rabernat: Ok then I will just bump keepbits to 11 for this test.

numcodecs/bitround.py
jakirkham: No we shouldn't be overwriting the input buffer.

observingClouds: How about changing
```
b = a.view(dtype=np.int32)
```
to
```
b = a.view(dtype=np.int32).copy()
```
to avoid the overwriting of the input buffer?

rabernat:
I want to apologize for letting this PR stall. I wanted to get it started as a proof of concept. If anyone else is excited about this feature and wants to work on it, I absolutely invite you to take over my PR and continue pushing it forward. cc @aaronspring

numcodecs/bitround.py
martindurant: `a.astype(np.int32, copy=True)` is more canonical.

martindurant:
It looks pretty complete - what more needs doing?

aaronspring:
I can add https://github.com/observingClouds/bitinformation_pipeline/blob/749623784286a04a54b6107fc8dd99d23d44f1b0/tests/test_bitround.py#L61 to testing, where I test this PR against Milan's bitinformation.round implementation in Julia.

martindurant:
> I can add

Does that require adding julia into the tests?
It would be nice to test for float64 as well as float32. Maybe float16 too, although I can't really imagine that being relevant.

aaronspring:
Sorry I meant I can add this to the discussion. Yes these tests run Julia code and shouldn’t be included here.

martindurant:
> Yes these tests run Julia code and shouldn’t be included here.

For a small test data array, we can inline the expected values.

numcodecs/bitround.py
```diff
@@ -0,0 +1,82 @@
...
+    "int64": np.float64
+}
+
+
+class BitRound(Codec):
+    """Floating-point bit rounding codec
+
+    Drops a specified number of bits from the floating point mantissa,
+    leaving an array more amenable to compression. The number of bits to keep should
+    be determined by a information analysis of the data to be compressed. See
```
rabernat, line 29:
```suggestion
    be determined by an information analysis of the data to be compressed. See 
```

that's my typo

martindurant, line 29:
```suggestion
    be determined by an information analysis of the data to be compressed. See
```

martindurant: (also removed trailing whitespace flagged by flake)

aaronspring:
When I compare this PRs [`round`](https://github.com/observingClouds/bitinformation_pipeline/blob/2f02df358521d28604b5be69ce9fb0388169e797/bitinformation_pipeline/bitround.py#L7) against `bitinformation.jl.round` in https://github.com/observingClouds/bitinformation_pipeline/blob/2f02df358521d28604b5be69ce9fb0388169e797/tests/test_bitround.py#L64, I get equals results only for `float32` but not for `float16` or `float64`, see https://github.com/observingClouds/bitinformation_pipeline/runs/5953867041?check_suite_focus=true

numcodecs/bitround.py
```diff
@@ -0,0 +1,82 @@
...
+        bits = max_bits[str(a.dtype)]
+        all_set = np.frombuffer(b"\xff" * a.dtype.itemsize, dtype=types[str(a.dtype)])
+        if self.keepbits == bits:
+            return a
+        if self.keepbits > bits:
+            raise ValueError("Keepbits too large for given dtype")
+        if not a.dtype.kind == "f" or a.dtype.itemsize > 8:
+            raise TypeError("Only float arrays (16-64bit) can be bit-rounded")
+        b = a.view(types[str(a.dtype)])
+        maskbits = 23 - self.keepbits
```
milankl, line 67: There's a hardcoded 23 here, which should be `maskbits = bits - self.keepbits` instead to allow `bits` to change: 10/23/52 for Float16/32/64

milankl (changes requested):

numcodecs/bitround.py
martindurant: Got it, thank you!

numcodecs/bitround.py
```diff
@@ -0,0 +1,82 @@
...
+            raise ValueError("keepbits must be zero or positive")
+        self.keepbits = keepbits
+
+    def encode(self, buf):
+        """Create int array by rounding floating-point data
+
+        The itemsize will be preserved, but the output should be much more
+        compressible.
+        """
+        a = ensure_ndarray(buf)
```
aaronspring, line 57: @martindurant can this also be implemented somehow differently? This line fails when using `xr.apply_ufunc(dask="allowed")` or `xr.map_blocks`, but it works `xr.apply_ufunc(dask="parallelized")` and eager data. Not critical but would be nice for scaling this function out
```
# bitinformation_pipeline.bitround._xr_bitround is round(data, keepbits) from test_bitround.py
xr.apply_ufunc(bitinformation_pipeline.bitround._xr_bitround, ds,7, dask='allowed')

---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Input In [6], in <cell line: 1>()
----> 1 xr.apply_ufunc(bitinformation_pipeline.bitround._xr_bitround, ds,7, dask='allowed')

File /work/mh0727/m300524/conda-envs/bitinfo/lib/python3.10/site-packages/xarray/core/computation.py:1159, in apply_ufunc(func, input_core_dims, output_core_dims, exclude_dims, vectorize, join, dataset_join, dataset_fill_value, keep_attrs, kwargs, dask, output_dtypes, output_sizes, meta, dask_gufunc_kwargs, *args)
   1157 # feed DataArray apply_variable_ufunc through apply_dataarray_vfunc
   1158 elif any(isinstance(a, DataArray) for a in args):
-> 1159     return apply_dataarray_vfunc(
   1160         variables_vfunc,
   1161         *args,
   1162         signature=signature,
   1163         join=join,
   1164         exclude_dims=exclude_dims,
   1165         keep_attrs=keep_attrs,
   1166     )
   1167 # feed Variables directly through apply_variable_ufunc
   1168 elif any(isinstance(a, Variable) for a in args):

File /work/mh0727/m300524/conda-envs/bitinfo/lib/python3.10/site-packages/xarray/core/computation.py:286, in apply_dataarray_vfunc(func, signature, join, exclude_dims, keep_attrs, *args)
    281 result_coords = build_output_coords(
    282     args, signature, exclude_dims, combine_attrs=keep_attrs
    283 )
    285 data_vars = [getattr(a, "variable", a) for a in args]
--> 286 result_var = func(*data_vars)
    288 if signature.num_outputs > 1:
    289     out = tuple(
    290         DataArray(variable, coords, name=name, fastpath=True)
    291         for variable, coords in zip(result_var, result_coords)
    292     )

File /work/mh0727/m300524/conda-envs/bitinfo/lib/python3.10/site-packages/xarray/core/computation.py:727, in apply_variable_ufunc(func, signature, exclude_dims, dask, output_dtypes, vectorize, keep_attrs, dask_gufunc_kwargs, *args)
    722     if vectorize:
    723         func = _vectorize(
    724             func, signature, output_dtypes=output_dtypes, exclude_dims=exclude_dims
    725         )
--> 727 result_data = func(*input_data)
    729 if signature.num_outputs == 1:
    730     result_data = (result_data,)

File ~/bitinformation_pipeline/bitinformation_pipeline/bitround.py:12, in _xr_bitround(data, keepbits)
     10 codec = BitRound(keepbits=keepbits)
     11 data = data.copy()  # otherwise overwrites the input
---> 12 encoded = codec.encode(data)
     13 return codec.decode(encoded)

File /work/mh0727/m300524/conda-envs/bitinfo/lib/python3.10/site-packages/numcodecs/bitround.py:57, in BitRound.encode(self, buf)
     51 def encode(self, buf):
     52     """Create int array by rounding floating-point data
     53 
     54     The itemsize will be preserved, but the output should be much more
     55     compressible.
     56     """
---> 57     a = ensure_ndarray(buf)
     58     bits = max_bits[str(a.dtype)]
     59     all_set = np.frombuffer(b"\xff" * a.dtype.itemsize, dtype=types[str(a.dtype)])

File /work/mh0727/m300524/conda-envs/bitinfo/lib/python3.10/site-packages/numcodecs/compat.py:45, in ensure_ndarray(buf)
     39     raise TypeError('array.array with char or unicode type is not supported')
     41 else:
     42 
     43     # N.B., first take a memoryview to make sure that we subsequently create a
     44     # numpy array from a memory buffer with no copy
---> 45     mem = memoryview(buf)
     47     # instantiate array from memoryview, ensures no copy
     48     arr = np.array(mem, copy=False)

TypeError: memoryview: a bytes-like object is required, not 'Array'
```

aaronspring: My usecase for this codec is to bitround input netcdf to output netcdf with xr.to_netcdf to disk, so this is a usecase outside of zarr

aaronspring: I do not understand the differences in all details, https://stackoverflow.com/questions/51736172/whats-the-difference-between-dask-parallelized-and-dask-allowed-in-xarrays-app

joshmoore: Sorry, I'm heading out the door. But at a glance, is this a #305 issue (if so, cc: @madsbk)

martindurant: @aaronspring , updated: does this work for you?

numcodecs/bitround.py
```diff
@@ -0,0 +1,82 @@
...
+
+        The itemsize will be preserved, but the output should be much more
+        compressible.
+        """
+        a = ensure_ndarray_like(buf)
+        if not a.dtype.kind == "f" or a.dtype.itemsize > 8:
+            raise TypeError("Only float arrays (16-64bit) can be bit-rounded")
+        bits = max_bits[str(a.dtype)]
+        all_set = np.frombuffer(b"\xff" * a.dtype.itemsize, dtype=types[str(a.dtype)])
+        if keepbits == bits:
```
aaronspring, line 62:
```suggestion
        if self.keepbits == bits:
```

aaronspring: same below

numcodecs/bitround.py
aaronspring: The change to `ensure_ndarray_like` still leads to the same problem unfortunately.
```
File /work/mh0727/m300524/conda-envs/bitinfo/lib/python3.10/site-packages/numcodecs/bitround.py:57, in BitRound.encode(self, buf)
     51 def encode(self, buf):
     52     """Create int array by rounding floating-point data
     53 
     54     The itemsize will be preserved, but the output should be much more
     55     compressible.
     56     """
---> 57     a = ensure_ndarray_like(buf)
     58     if not a.dtype.kind == "f" or a.dtype.itemsize > 8:
     59         raise TypeError("Only float arrays (16-64bit) can be bit-rounded")

File /work/mh0727/m300524/conda-envs/bitinfo/lib/python3.10/site-packages/numcodecs/compat.py:42, in ensure_ndarray_like(buf)
     38     raise TypeError("array.array with char or unicode type is not supported")
     39 else:
     40     # N.B., first take a memoryview to make sure that we subsequently create a
     41     # numpy array from a memory buffer with no copy
---> 42     mem = memoryview(buf)
     43     # instantiate array from memoryview, ensures no copy
     44     buf = np.array(mem, copy=False)

TypeError: memoryview: a bytes-like object is required, not 'DataArray'
```

martindurant: Hm, so not that. Can you try with https://github.com/zarr-developers/numcodecs/pull/299/commits/9622fe318777d769d170f650eaa304dd5c6e5b3b (i.e., don't call `encode()`, call the `_bitround()` function)

aaronspring: Oh, but it works with `dask.array.map_blocks` when `import dask.array as da; da.map_blocks(bitinformation_pipeline.bitround._np_bitround, ds[v].data, keepbits=7, dtype="float32").compute()`, likely I use `xr.map_blocks` in a wrong way

aaronspring: happens in both `encode()` and `decode()`

aaronspring: but as `dask.array.map_blocks` works, I will open an issue in `xarray`. And I am very happy that `xr.apply_ufunc(dask="parallelized")` works out-of-the-box!

martindurant: What is `bitinformation_pipeline`?

aaronspring: we are wrapping @milankl's `bitinformation.jl` to be called in `python` and directly apply these functions to `xarray` objects.

aaronspring: > Oh, but it works with `dask.array.map_blocks` when `import dask.array as da; da.map_blocks(bitinformation_pipeline.bitround._np_bitround, ds[v].data, keepbits=7, dtype="float32").compute()`, likely I use `xr.map_blocks` in a wrong way

started https://github.com/pydata/xarray/discussions/6482

martindurant: Did you try my `_bitround()` version of this?

aaronspring: Yes, it uses this PR. https://github.com/pydata/xarray/discussions/6482 and the docs explain that xr.apply_ufunc(dask="allowed") doesnt work for numpy functions.

aaronspring: sorry, my fault dragging you into this...

martindurant: OK, let's mark this resolved then - I'll say that the current state of the code is good.

numcodecs/bitround.py
```diff
@@ -0,0 +1,82 @@
...
+        b &= mask
+        return b
+
+    def decode(self, buf, out=None):
+        """Remake floats from ints
+
+        As with ``encode``, preserves itemsize.
+        """
+        buf = ensure_ndarray_like(buf)
+        dt = buf.dtype if buf.dtype.kind == "f" else inverse[str(buf.dtype)]
```
jakirkham, line 80:
```suggestion
        # Cast back from `int` to `float` type (if a `float`ing type buffer is provided)
        dt = buf.dtype if buf.dtype.kind == "f" else np.dtype(buf.dtype.str.replace("i", "f"))
```

numcodecs/bitround.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+from .abc import Codec
+from .compat import ensure_ndarray_like, ndarray_copy
+
+max_bits = {
+    "float16": 10,
+    "float32": 23,
+    "float64": 52,
+}
```
jakirkham, line 11: Should we have some comment (or module docstring) explaining what these are. The `max_bits` is less straightforward (at least to me). So a bit more detail on that would be helpful.

numcodecs/bitround.py
```diff
@@ -0,0 +1,82 @@
...
+types = {
+    "float16": np.int16,
+    "float32": np.int32,
+    "float64": np.int64,
+}
+inverse = {
+    "int16": np.float16,
+    "int32": np.float32,
+    "int64": np.float64
+}
```
jakirkham, line 21: We might be able to drop these with other code handling. See other comments.

jakirkham (commented):
Thanks for continuing to push this forward Ryan and Martin! 😄

Sorry for being somewhat absent here. Though am happy to see others providing you feedback to improve on this work.

Had a couple minor comments below. Not attached to any particular code suggestions, but thinking we can do a bit of simplification in a few spots and provide a bit more context for future readers. Happy to clarify anything if needed 🙂

numcodecs/bitround.py
```diff
@@ -0,0 +1,82 @@
...
+        """Create int array by rounding floating-point data
+
+        The itemsize will be preserved, but the output should be much more
+        compressible.
+        """
+        a = ensure_ndarray_like(buf)
+        if not a.dtype.kind == "f" or a.dtype.itemsize > 8:
+            raise TypeError("Only float arrays (16-64bit) can be bit-rounded")
+        bits = max_bits[str(a.dtype)]
+        all_set = np.frombuffer(b"\xff" * a.dtype.itemsize, dtype=types[str(a.dtype)])
```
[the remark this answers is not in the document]
jakirkham: Note this is a no-op for `int` types, which may be handy.

Didn't give any thought to other types which are passed through as is. Maybe we could make them all `int`s like so?

```suggestion
        a_int_dtype = np.dtype(f"{dt.byteorder}i{dt.itemsize}")
        all_set = np.frombuffer(b"\xff" * a.dtype.itemsize, dtype=a_int_dtype)
```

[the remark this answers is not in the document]
martindurant: We do require `a.dtype.kind == "f"`, a few lines above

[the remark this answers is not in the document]
jakirkham: Sure Ryan mentioned handling more types. So was thinking about what that would look like.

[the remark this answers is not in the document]
martindurant: I think it's fair to say that this is for float only (of various sizes). Only complex would make sense in addition, but I'm prepared to ignore that for now.

numcodecs/bitround.py
```diff
@@ -0,0 +1,82 @@
...
+        depends on the dtype input data. If keepbits is
+        equal to the maximum allowed for the data type, this is equivalent
+        to no transform.
+    """
+
+    codec_id = 'bitround'
+
+    def __init__(self, keepbits: int):
+        if keepbits < 0:
+            raise ValueError("keepbits must be zero or positive")
```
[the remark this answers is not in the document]
martindurant: I guess you get a constant back, so you only "keep" the exponent information, like rounding to nearest factor of 2. Still allowed.

martindurant:
@jakirkham , I agree and adopted your changes.

numcodecs/bitround.py
```diff
@@ -0,0 +1,83 @@
...
+max_bits = {
+    "float16": 10,
+    "float32": 23,
+    "float64": 52,
+}
+types = {
+    "float16": np.int16,
+    "float32": np.int32,
+    "float64": np.int64,
+}
```
jakirkham, line 19: Is this still needed? Appears unused below

```suggestion
```

martindurant:
Do we need to drop py36? Something's up with the environment.
cf #308 (for adding 3.10)

martindurant:
@jakirkham , any thoughts?

aaronspring:
I think we can now rerun CI after merging from main

joshmoore:
Fixed the release conflicts.

rsignell-usgs:
@rabernat, should I be able to just specify this in the `encoding` argument dict in `ds.to_zarr()`?
It is a filter, right?

aaronspring:
See example in https://github.com/observingClouds/xbitinfo/issues/75

rsignell-usgs:
@aaronspring , awesome!   That's exactly what I was looking for!

martindurant:
Will merge at the end of my day if there are no more comments.

numcodecs/bitround.py
```diff
@@ -0,0 +1,78 @@
...
+
+class BitRound(Codec):
+    """Floating-point bit rounding codec
+
+    Drops a specified number of bits from the floating point mantissa,
+    leaving an array more amenable to compression. The number of bits to keep should
+    be determined by an information analysis of the data to be compressed. See
+    https://github.com/zarr-developers/numcodecs/issues/298 for discussion
+    and the original implementation in Julia referred to at
+    https://www.nature.com/articles/s43588-021-00156-2
```
rabernat, line 25:
```suggestion
    https://github.com/milankl/BitInformation.jl
```

numcodecs/bitround.py
```diff
@@ -0,0 +1,78 @@
...
+    "float64": 52,
+}
+
+
+class BitRound(Codec):
+    """Floating-point bit rounding codec
+
+    Drops a specified number of bits from the floating point mantissa,
+    leaving an array more amenable to compression. The number of bits to keep should
+    be determined by an information analysis of the data to be compressed. See
```
rabernat, line 22:
```suggestion
    be determined by an information analysis of the data to be compressed.
    The approach is based on the paper by Klöwer et al. 2021
    (https://www.nature.com/articles/s43588-021-00156-2). See
```

rabernat (commented):
Just one minor docstring suggestion.

Thanks for seeing this through Martin!

jakirkham (approved):

jakirkham:
Thanks Martin & Ryan and everyone who helped review! 🙏

milankl:
There’s a hardcoded 23 (representing the number of mantissa bits) in line 67, that should be 10,23,52 for float16/32/64. So that line should be ‘maskbits = bits - self.keepbits’ I guess

milankl:
Sorry, I don't know why this post was sent now. I may have sent an email at the beginning of the year to contribute when the PR was still open, maybe this email got lost and only delivered now? Weird. The PR clearly contains the `maskbits = bits - self.keepbits` lines instead of a hardcoded 23, so consider my previous comment as irrelevant.

martindurant:
OK then! I'll blame the gremlins :)

numcodecs/bitround.py
jakirkham: Continued in PR: https://github.com/zarr-developers/numcodecs/pull/608

Issue #308, opened 2022-02-17
release wheel for 3.10


ianhi:
#### Minimal, reproducible code sample, a copy-pastable example if possible
```
mamba create -n test python=3.10
pip install numcodecs
```

#### Problem description
Above code spends some time building the wheel. Which may fail on some systems and also slows down install times. It would be great to bump CI python versions and release the new wheels.

#### Version and installation information

Please provide the following:

* Value of ``numcodecs.__version__`` 0.9.1
* Version of Python interpreter 3.10
* Operating system (Linux/Windows/Mac) linux
* How NumCodecs was installed (e.g., "using pip into virtual environment", or "using conda"): pip

martindurant:
I think conda-forge has built for 3.10 already, if that will do.

ianhi:
> I think conda-forge has built for 3.10 already, if that will do.

That's a good workaround - but I expect that there will be people who use pip primarily  who will also run into this. e.g. i ran into this when doing pip install of different pacakge and it caught me off gaurd. No rush from me to fix, just wanted to bring it up

martindurant:
Sure, we need to update the release too.

jakirkham:
We build wheels with [`cibuildwheel` in this GH Action]( https://github.com/zarr-developers/numcodecs/blob/master/.github/workflows/wheel.yaml ). AFAICT `cibuildwheel` already supports Python 3.10. So that should just work.

Currently CI doesn't [cover Python 3.10]( https://github.com/zarr-developers/numcodecs/blob/a81856efbf8dfd8ea925acaac01235a096d285e4/.github/workflows/ci-linux.yaml#L11 ). So adding that would be good to do.

If someone would like to do a PR to add Python 3.10, that would be a good first step towards addressing this. Not expecting any major issues there, but if they come happy to discuss. @ianhi would you be interested in giving this a try? 🙂

martindurant:
> cibuildwheel already supports Python 3.10

Yes, only if it is up to date

henryiii:
[...]

PR #309, opened 2022-02-17
chore: bump cibuildwheel version, use action

henryiii:
Followup to #308. General recommendations:

* There should be a pyproject.toml. Then these cibuildwheel settings can be put there. You also need the lines here if you add one: https://scikit-hep.org/developer/packaging#pep-517518-support-high-priority - FYI, `setup_requires` is deprecated and produces a warning.
* Checkout and upload/download actions are using v1. You should use v2.
* You can set up a dependabot update following https://cibuildwheel.readthedocs.io/en/stable/faq/#automatic-updates
* This will also enable musllinux wheels, can be manually skipped if you don't want to ship for musl
* This will use manylinux2014 as default for all wheels - can be set lower if you need manylinux2010 still (ideally in TOML, so you can set <3.9 as 2010 and 3.10+ as 2014 using overrides). Looks like manylinux2010 is going away, though: https://github.com/pypa/manylinux/issues/1281. Mostly 3.6 & 3.7 are affected. Dropping before that happens is one solution. ;)
* It doesn't look like you are building for M1, see https://cibuildwheel.readthedocs.io/en/stable/faq/#apple-silicon (scroll down a little for an example). There's also guides to build for emulated linux architectures.

I'd also recommend testing on 3.10. You could add 3.10 and drop 3.6, if you want to keep to 4 versions (or non-EOL) versions supported.

### DIFF
--- a/.github/workflows/wheel.yaml
+++ b/.github/workflows/wheel.yaml
@@ -23,7 +23,7 @@ jobs:
           submodules: true
 
       - uses: pypa/cibuildwheel@v2.3.1
-        
+
       - uses: actions/upload-artifact@v1
         with:
           name: wheels
--- a/.github/workflows/ci-linux.yaml
+++ b/.github/workflows/ci-linux.yaml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.6", "3.7", "3.8", "3.9"]
+        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10"]
         os: [ubuntu-18.04, ubuntu-20.04]
 
     steps:
--- a/.github/workflows/ci-osx.yaml
+++ b/.github/workflows/ci-osx.yaml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.6", "3.7", "3.8", "3.9"]
+        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10"]
 
     steps:
       - name: Checkout source
--- a/.github/workflows/ci-windows.yaml
+++ b/.github/workflows/ci-windows.yaml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.6", "3.7", "3.8", "3.9"]
+        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10"]
 
     steps:
       - name: Checkout source


.github/workflows/wheel.yaml
```diff
@@ -22,18 +22,8 @@ jobs:
...
-          python-version: '3.9'
-
-      - name: Install cibuildwheel
-        run: |
-          python -m pip install cibuildwheel==1.8.0 wheel
-
-      - name: Build wheel
-        run: |
-          python -m cibuildwheel --output-dir wheelhouse
+      - uses: pypa/cibuildwheel@v2.3.1
```
jakirkham, line 25: Really like this simplification 😄

.github/workflows/wheel.yaml
```diff
@@ -53,7 +43,7 @@ jobs:
           python-version: '3.9'
 
       - name: Build sdist
-        run: python setup.py sdist
+        run: pipx run build --sdist
```
jakirkham, line 46: Do we need to add `pipx` as a requirement somewhere? Or is that already installed?

.github/workflows/wheel.yaml
```diff
@@ -13,7 +13,7 @@ jobs:
     env:
       CIBW_TEST_COMMAND: pytest --pyargs numcodecs
       CIBW_TEST_REQUIRES: pytest
-      CIBW_SKIP: "*27* pp* *35*"
+      CIBW_SKIP: "pp*"
```
jakirkham, line 16: AFAICT cibuildwheel still supports Python 3.6, correct? Is there anyway for us to state here that we still support it? Just want to make sure if an update to cibuildwheel drops Python 3.6 from cibuildwheel that we catch it (admittedly maybe we should be looking to drop Python 3.6 anyways just want to make sure we do that explicitly as opposed to implicitly due to tooling changes).

jakirkham (commented):
Thanks Henry! This is great 😄

Had a couple questions below 🙂

.github/workflows/wheel.yaml
henryiii: It's built into the default runners as a supported package manager. See https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-Readme.md or any other platform under https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-software . It does not require setup-python (that could be moved below this line). (pipx is included in manylinux images, too, which is really handy when you want to run something like nox/tox on many Pythons ;) )

.github/workflows/wheel.yaml
jakirkham: Ok sounds like there isn't a good way to do this, but that's ok since Numcodecs will most likely drop Python 3.6 before cibuildwheel does.

jakirkham (approved):
Thanks again Henry! LGTM 😄

jakirkham:
@joshmoore would you be able to review? 🙂

joshmoore:
> would you be able to review?

LGTM as well. We're currently at `v0.10.0.a1`. Happy to trigger an `a2` to test the process. Then we should probably start considering a release with the Entrypoints PR.

joshmoore:
In that case, let me try it _here_.

joshmoore:
All the new 3.10 builds passed. Just waiting on the slower wheels builds now.

jakirkham:
Looks green. Sounds like we are happy here. So went ahead and merged. Hope that is ok 🙂

Thanks Henry and Josh! 😀

joshmoore:
v0.10.0.a2 pushed pushed

PR #311, opened 2022-02-25
Fix macOS Python 3.10

jakirkham:
For some reason we pick up `clang` version `12.0.0` only in the Python 3.10 job, which has problems with TAPI (hence the CI issues we have been seeing). These issues are not present with `clang` version `12.0.1`+ (hence why the other Python jobs work). So here we require `clang` version `12.0.1`+ when installing compilers to fix CI.

Edit: Also included a bit more debug info in CI to aid with issue templates that Conda & conda-forge use

TODO:

### DIFF
--- a/.github/workflows/ci-osx.yaml
+++ b/.github/workflows/ci-osx.yaml
@@ -24,6 +24,13 @@ jobs:
         env:
             ACTIONS_ALLOW_UNSECURE_COMMANDS: true
 
+      - name: Show info about `base` environment
+        shell: "bash -l {0}"
+        run: |
+          conda info
+          conda config --show-sources
+          conda list --show-channel-urls
+
       - name: Set up env
         shell: "bash -l {0}"
         run: |
@@ -32,6 +39,11 @@ jobs:
           which pip
           pip install -r requirements_test.txt -r requirements.txt
           conda env export
+
+      - name: Show info about `env` environment
+        shell: "bash -l {0}"
+        run: |
+          conda list --show-channel-urls -n env
 
       - name: Install numcodecs
         shell: "bash -l {0}"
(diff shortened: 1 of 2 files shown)


jakirkham:
cc @joshmoore @rabernat (for vis)

joshmoore:
Odd, but :+1:

jakirkham:
Yeah I filed issue ( https://github.com/conda-forge/clangdev-feedstock/issues/166 ) about this

.github/workflows/ci-osx.yaml
```diff
@@ -34,7 +34,7 @@ jobs:
       - name: Set up env
         shell: "bash -l {0}"
         run: |
-          conda create -n env python==${{matrix.python-version}} wheel pip compilers 'clang>=12.0.1'
+          conda create -n env python=${{matrix.python-version}} wheel pip compilers 'clang>=12.0.1'
```
jakirkham, line 37: Having `==`s effectively means `3.10.0` instead of `3.10.0`+. As we likely don't want to pin patch releases to `0`, relaxing that here. May be part of the issue we were seeing as well

joshmoore: Oh wow. TIL. :+1:

jakirkham:
Was unable to reproduce the issue locally.

Have `conda` version `4.11.0` (same as CI). A slightly newer version of `python` is used in the `base` environment (both locally and on CI it is 3.9) and a newer macOS, but doubt either of those is causing the solver issue.

In any event this should unblock CI here. Will see if anything turns up in the upstream issue.

jakirkham:
CI looks green ✅

jakirkham:
Merging to fix CI. Hope that is ok.

Issue #312, opened 2022-02-25
Run wheel builds selectively?


jakirkham:
Given the wheel CI builds take a while, wonder if we could limit how often we run these. For example:

* Only run if certain files (like `setup.py` change)
* Run only when requested (like with a PR label)
* Run only on `main` and/or tags
* Test some subset of wheel builds and only test all in specific contexts
* ?

We need not do all or any of these. They are just some thoughts.

Curious what people think 🙂

jakirkham:
cc @henryiii (in case you have thoughts on what others do typically or on ways we might speed up the builds)

joshmoore:
Another option is cron, but for my part I'll add that I'm not great at keeping up with non-PR action failures. (And cronjobs get disabled after 6(?) months with no actions)

jakirkham:
Yeah and issues can pile up if not run regularly

Just trying to find a balance since it seems atm we are waiting 2hrs for the Linux wheel build, which is kind of long

henryiii:
I usually only run via `workflow_dispatch` or {github releases|tags} (depending on the project - I like releases, but sometimes tags are needed or better). For main / stable workflows, I'll have wheel builds on stable only. I'll sometimes add a very small subset of runs (one wheel each on each OS, for example) into the normal testing workflow. Generally wheel building is pretty expensive to serve as your main testing, traditional testing is better - (full) wheel builds should target releases. For example, see https://github.com/scikit-hep/boost-histogram/blob/339306d388d7f112afb4e6f619c52ab8252c4f6a/.github/workflows/tests.yml#L89-L114.

henryiii:
(this is also another reason to use configuration like https://github.com/scikit-hep/boost-histogram/blob/339306d388d7f112afb4e6f619c52ab8252c4f6a/pyproject.toml#L74-L86 instead of putting everything in environment variables)

jakirkham:
Thanks for the pointers here Henry! 😄

Looking more at the builds here. It seems we are covering other architectures that probably require emulation (like `ppc64le` and `aarch64`) in CI, 32-bit builds (Idk that we have tested there on CI or otherwise outside of the wheel builds) as well as other `libc` implementations (like `musl`).

[...]

PR #315, opened 2022-04-12
ci: Build linux aarch64 wheels on release

jthetzel:
Testing merge conflicts from https://github.com/zarr-developers/numcodecs/pull/289 .


jakirkham:
Thanks for exploring this Jeremey! 😄

Generally supportive of helping more users on more architectures. Though it is worth noting we are struggling with CI burden ( https://github.com/zarr-developers/numcodecs/issues/312 ). Would suggest that we start trimming what runs on every PR in favor of running some things only on `main` or even only on release.

jakirkham:
cc @joshmoore

jthetzel:
`Wheels / Build wheel on auto for macos-latest (pull_request)` complaining with
```
ERROR: No matching distribution found for cibuildwheel==1.8.0
```

.github/workflows/wheel.yaml
```diff
@@ -24,6 +29,26 @@ jobs:
 
       - uses: pypa/cibuildwheel@v2.3.1
 
+      - uses: docker/setup-qemu-action@v1
+        if: ${{ matrix.arch == 'aarch64' }}
+        name: Set up QEMU
+
+      - name: Install cibuildwheel
+        run: |
+          python -m pip install cibuildwheel==1.8.0 wheel
```
jakirkham, line 38: Do we need this pinning or could this be dropped?

rabernat:
It would be great to get these wheels out.  @jthetzel and @jakirkham - in your opinion, what's needed to move forward?

jakirkham:
As noted above ( https://github.com/zarr-developers/numcodecs/pull/315#issuecomment-1097131621 ), am concerned about CI usage. Currently we run wheel builds on every PR regardless of what the PR changes. These builds take quite a while to run. We've cutdown the number of builds done in wheel jobs to help mitigate this. However adding new architectures that require emulation (like ARM) will make these wheel jobs take longer. So think we need to come up with a way to only selectively enable these jobs. Please see issue ( https://github.com/zarr-developers/numcodecs/issues/312 ) for more details.

Separate from that CI here is failing. Don't have time to look into this unfortunately. So cannot advise on that atm.

keflavich:
I have the same issue, and while I can use conda for the install to _use_ the code, it prevents me from _testing_ with `tox`, since my tox environment is set up to use pip.   I'm not sure how to read this - have you guys decided not to build arm64 wheels because of the CI burden?  Or are you just going to limit the versions that you build wheels for?

jakirkham:
I don't think we are against building the wheels. Just as this PR is configured currently, it builds them on every PR, which adds up to a lot of CI usage and negatively impacts the contributor experience. If we could do this only when there is a release and/or only when there are changes to build related files (`setup.py`, `pyproject.toml`, etc.), that would help address this concern.

jthetzel:
The continuous integration is updated to only build aarch64 wheels when a release is created (when a tag matching "v*" is created).

jthetzel:
Building aarch64 wheels for python 3.8, 3.9, 3.10, 3.11 takes about 1.5 hours ( https://github.com/jthetzel/numcodecs/actions/runs/5500370791 ).

jthetzel:
@jakirkham @rabernat I completely forgot about this pull request until I went to build a container on a jetson nano with a zarr dependency this morning, googled my way here, and was shocked that I contributed to it over a year ago (my daughter was born five days later).

Having a separate aarch64 wheel build action that only triggers when a release is deployed feels clean to me. Until aarch64 build vms are widely available, I'm not aware of other options, but intereseted in further discussion.

jthetzel:
@joshmoore Bumping this just to check if more work is needed or a different path preferred.

joshmoore:
Thanks for the ping, @jthetzel. I've relaunched the builds. I'm on the road though and so will be a bit slow on the uptake until next week. cc: @jakirkham

Did you happen to see https://github.com/zarr-developers/numcodecs/pull/428 for a related / alternative proposal?

dcherian:
Can this be merged? It'd be nice to have `aarch64` wheels

dstansby:
Instead of adding a whole new GitHub actions workflow, I think it's easier to add support for linux aarch64 to the existing wheel workflow: https://github.com/zarr-developers/numcodecs/blob/main/.github/workflows/wheel.yaml

joshmoore:
@jthetzel: are you interested in giving that a try?

jthetzel:
Certainly. I'll update the branch an re-read the discussion.

jakirkham:
The main reason we want to split them out is CI time blows up if we have them all in one job

Perhaps another approach is we only build these on `main` or perhaps at release time

dstansby:
> The main reason we want to split them out is CI time blows up if we have them all in one job

Do you mean split them out into multiple GitHub actions jobs that run in parallel? If so the current approach to having a matrix across operating systems achieves this in a [single configuration file](https://github.com/zarr-developers/numcodecs/blob/main/.github/workflows/wheel.yaml), which produces parallel workflows for each OS (e.g. see https://github.com/zarr-developers/numcodecs/actions/runs/10338358292).

So I think to minimise duplicating code/GH actions config, the way to go is to update the existing configuration file instead of adding a new one?

agriyakhetarpal:
This PR was superseded by #710. I guess it can be closed, and I apologise for taking over the work done here; I didn't see this effort at all until I opened my PR as it wasn't linked to be closing the original issue 🙈

PR #318, opened 2022-05-05
Drop Python 3.6

joshmoore:
see: #317
see: https://github.com/zarr-developers/zarr-python/issues/765

- [ ] Unit tests and/or doctests in docstrings
- [ ] `tox -e py39` passes locally
- [ ] Docstrings and API docs for any new/modified user-facing classes and functions
- [ ] Changes documented in docs/release.rst
- [ ] `tox -e docs` passes locally
- [ ] GitHub Actions CI passes
- [ ] Test coverage to 100% (Coveralls passes)

PR #156, opened 2018-12-03
Drop win32 support

alimanfoo:
Removes 32-bit appveyor builds. Resolves #97.

TODO:
* [ ] Unit tests and/or doctests in docstrings
* [ ] ``tox -e py37`` passes locally
* [ ] ``tox -e py27`` passes locally
* [ ] Changes documented in docs/release.rst
* [ ] ``tox -e docs`` passes locally
* [ ] AppVeyor and Travis CI passes
* [ ] Test coverage to 100% (Coveralls passes)

PR #320, opened 2022-05-06
Trim wheel builds

jakirkham:
Partially addresses issue ( https://github.com/zarr-developers/numcodecs/issues/312 )

The wheel build matrix is taking [2.5hrs to go through]( https://github.com/zarr-developers/numcodecs/runs/5668831517?check_suite_focus=true ). Some of these builds are for novel architectures or deployments that are not really tested. Trim the build matrix to a more reasonable set of build to get the build time more manageable.

TODO:

### DIFF
--- a/.github/workflows/wheel.yaml
+++ b/.github/workflows/wheel.yaml
@@ -13,7 +13,7 @@ jobs:
     env:
       CIBW_TEST_COMMAND: pytest --pyargs numcodecs
       CIBW_TEST_REQUIRES: pytest
-      CIBW_SKIP: "pp* cp36-*"
+      CIBW_SKIP: "pp* cp36-* *win32 *_i686"
       CIBW_ENVIRONMENT: "DISABLE_NUMCODECS_AVX2=1"
       CIBW_ENVIRONMENT_MACOS: 'MACOSX_DEPLOYMENT_TARGET=10.9 DISABLE_NUMCODECS_AVX2=1 CFLAGS="$CFLAGS -Wno-implicit-function-declaration"'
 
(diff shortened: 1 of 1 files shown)


jakirkham:
Am still seeing `build (3.6)` in the statuses above even though PR ( https://github.com/zarr-developers/numcodecs/pull/318 ) was merged. Must be missing something still

cc @joshmoore

jakirkham:
Much better. The longest running of these is [macOS at 14mins]( https://github.com/zarr-developers/numcodecs/runs/6327719159?check_suite_focus=true ).

joshmoore (approved):
The change itself (i.e. syntactically) looks fine, but I don't have any particular thoughts on the potential coverage. Would running these on tag be a compromise?

jakirkham:
Well the question from my perspective is who is asking for these other builds? Currently we have people asking about ARM in various cases, which can be good to include. There are often PPC users at HPC facilities. So that can matter. Neither of these is currently excluded, but we may need to do additional work to add them (ideally in a separate matrix item). They will also take additional resources.

However I'm not aware of folks asking for musl support or s390x. These are actually eating up a fair amount of time to build (this includes emulation cost for s390x) despite not having any demand AFAICT.

Also 32-bit was dropped a long time ago ( https://github.com/zarr-developers/numcodecs/pull/156 ). So it is kind of weird that it is coming back through wheels when it isn't supported.

To summarize, the things being dropped here are things AFAIK no one has asked for and/or are not supported. The result is all PRs taking ~2.5hrs to run on CI, which is pretty significant friction for the average contributor. I don't have a problem adding and filtering new things as users request them, but spending a lot of time building things nobody is asking us for or we don't support doesn't make a lot of sense (at least not to me anyways).

joshmoore:
All fair. Worst case they'll let us know on twitter. :wink:

jakirkham:
Thanks Josh 😀

And if they do happy to revisit then 😉

Issue #319, opened 2022-05-06
Drop/replace `array_decode_utf8` from `benchmark_vlen.ipynb`


jakirkham:
In PR ( https://github.com/dask/fastparquet/pull/580 ) included in `fastparquet` release `0.6.0`, `array_decode_utf8` was dropped from `fastparquet.speedups`. We should either drop or replace [this benchmark from `notebook/benchmark_vlen.ipynb`]( https://github.com/zarr-developers/numcodecs/blame/15f17138b9b41e99faa4a2670e2580a782547d6b/notebooks/benchmark_vlen.ipynb#L45-L67 ).

For context this notebook was originally added in PR ( https://github.com/zarr-developers/numcodecs/pull/56 ).

jakirkham:
cc @martindurant (in case you have any thoughts here; no worries if not 🙂)

martindurant:
I had no idea that this existed!

jakirkham:
The function or the benchmark? In either case think it was totally reasonable to drop that code from `fastparquet` (it looked kind of like an internal function). Just didn't know if there was something else we should be using here or if we should just drop it.

martindurant:
The encode part was rolled into pack_byte_array, so note reachable from outside, definitely an internal thing.

PR #321, opened 2022-05-06
Drop `fastparquet` benchmark

jakirkham:
Fixes https://github.com/zarr-developers/numcodecs/issues/319

The `array_decode_utf8` function has gone away in `fastparquet` version `0.6.0`. So the `FastParquetCodec` created here no longer works. There doesn't seem to be a good alternative function to use in its place. Given this, go ahead and drop `fastparquet` from the benchmark.

TODO:

### DIFF
--- a/notebooks/benchmark_vlen.ipynb
+++ b/notebooks/benchmark_vlen.ipynb
@@ -42,37 +42,6 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "import fastparquet\n",
-    "\n",
-    "\n",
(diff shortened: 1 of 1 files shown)


martindurant:
+1

Makes me wonder whether the old code was actually fast of not :)

jakirkham:
Well the numbers currently in there are from me rerunning today with `0.5.0` on Python 3.9. It seems pretty close to `VLenUTF8` included in Numcodecs, which is doing a similar thing.

The results from when Alistair ran it were pretty similar as well.

PR #332, opened 2022-07-05
Add entrypoints to setup.py

joshmoore:
0.10.0 has broken conda-forge builds. Propose for a quick 0.10.1 release.

see:
 - https://github.com/conda-forge/numcodecs-feedstock/pull/86
 - https://github.com/zarr-developers/numcodecs/pull/300
 - https://github.com/conda-forge/numcodecs-feedstock/issues/82

TODO:


joshmoore:
Looks like the tests aren't passing on Windows, @martindurant.

```
  ==================================== ERRORS ====================================
  __________________ ERROR at teardown of test_entrypoint_codec __________________
  
      @pytest.fixture()
      def set_path():
          sys.path.append(here)
          numcodecs.registry.run_entrypoints()
          yield
          sys.path.remove(here)
          numcodecs.registry.run_entrypoints()
  >       numcodecs.registry.codec_registry.pop("test")
  E       KeyError: 'test'
  
  /tmp/tmp.WpzI00cMS0/venv/lib/python3.7/site-packages/numcodecs/tests/test_entrypoints.py:20: KeyError
```

martindurant:
I guess we can always `pop("test", None)`, but why isn't it there in the first place? Perhaps this test would do well to interrogate the state of `numcodecs.registry.entries`.

martindurant:
I just can't get it to fail :|
With this very blanch and a new fresh env and steps as given in the ci script; all works fine. Is there a new version of `entrypoints`?

joshmoore:
`wheel.yml` doesn't look to use any of the requirements files so should be only using `setup.py`.

setup.py
```diff
@@ -370,6 +370,7 @@ def run_setup(with_extensions):
         maintainer_email='alimanfoo@googlemail.com',
         url='https://github.com/zarr-developers/numcodecs',
         license='MIT',
+        zip_safe=False,
```
jakirkham, line 373: This might not be needed. Just trying something 🍀

That said, we probably already should have had this given we have shared libraries

Or at least shared libraries misbehaving was what had usually prompted this before. Maybe it is [not as relevant anymore]( https://setuptools.pypa.io/en/latest/deprecated/zip_safe.html )

martindurant:
You define your own tests in conda-forge. Some packages just test that the imports work, others run the whole test suite.

martindurant:
Should we, in the end, ignore entrypoints and use builtins only? Like the following:
```python
out = []
for dist in importlib.metadata.distributions():
    eps = [ep for ep in dist.entry_points if ep.group == "numcodecs.codecs"]
    if eps:
        out.extend(eps)
```

This takes 0.1s on my bloated env, and we only do it once, all taken up by path-string manipulating and config file parsing; however, nothing is cached, so other packages that do this kind of thing (e.g., intake) would end up repeating the work.

martindurant:
Well, OK; but it would be good to know if any real situations face this problem! I suppose we may start to see reports as people use entrypoints-based codecs, perhaps via kerchunk.

joshmoore:
Definitely! But the current situation is breaking downstream already.

martindurant:
Sounds like you should hit the big green button, @joshmoore

jakirkham:
It might be worth configuring automerge and/or disabling the requirement to update branches to latest. Just a thought 🙂

joshmoore:
Created https://github.com/zarr-developers/numcodecs/issues/343

Merging.

PR #356, opened 2022-10-05
FixedScaleOffset doc: change scale parameter type to float

bogovicj:
This PR is a small change to the documentation of the FixedScaleOffset:
changing the type of the scale parameter from `int` to `float`.

The filter currently behaves reasonably and as I would expect if a passing a float to the scale parameter:

```python
# random angles in the range [-pi,pi]
x = (np.random.random([64]) * 2 * np.pi) - np.pi

# encode in the range [0,255] with a byte
f = FixedScaleOffset(scale= (255.0 / (2.0 * np.pi) ), offset=-np.pi, dtype='float64', astype='uint8')

# behaves as expected
x_enc = f.encode(x)
x_dec = f.decode(x_enc)
```

and is a potentially useful application of this filter.  

As well, this documentation change makes this filter's description consistent with the  [`scale_factor` and `add_offset` attributes of NetCDF](https://docs.unidata.ucar.edu/netcdf-c/current/attribute_conventions.html).

> When scale_factor and add_offset are used for packing, the associated variable (containing the packed data) is typically of type byte or short, whereas the unpacked values are intended to be of type float or double. **The attributes scale_factor and add_offset should both be of the type intended for the unpacked data, e.g. float or double.**

(emphasis mine) 

TODO:


jakirkham:
Thanks John! 🙏

This makes sense to me. Should we add an example in the docstring and maybe [a test filter here]( https://github.com/zarr-developers/numcodecs/blob/5230f03a04f2791e3e45c992a96d8871f57dc96e/numcodecs/tests/test_fixedscaleoffset.py#L23-L29 )?

dstansby (approved):
Since this is an improvement I think worth getting it in, even without an example.

jakirkham:
Adding [a test to this file]( https://github.com/zarr-developers/numcodecs/blob/2a7bf9cdfb1ec3861ac79a9ace10201a2a032272/numcodecs/tests/test_fixedscaleoffset.py ) would be good though. As currently this is untested behavior

PR #365, opened 2022-10-22
Update developer installation guide

Saransh-cpp:
- use `--recursive` to clone the submodules too (I get a build error locally if `c-blosc` is not cloned)
- use the generic `HTTPS` URLs instead of the `SSH` ones (not everyone has `SSH` keys)
- specify a command to install `numcodecs` globally (I noticed that running `python setup.py build_ext --inplace` did not install `numcodecs` in my `python` environment, but as it correctly built the package, the tests passed)

TODO:

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -75,9 +75,9 @@ The NumCodecs source code is hosted on GitHub at the following location:
 You will need your own fork to work on the code. Go to the link above and hit
 the "Fork" button. Then clone your fork to your local machine::
 
-    $ git clone --recursive https://github.com/your-user-name/numcodecs.git
+    $ git clone --recursive git@github.com:your-user-name/numcodecs.git
     $ cd numcodecs
-    $ git remote add upstream https://github.com/zarr-developers/numcodecs.git
+    $ git remote add upstream git@github.com:zarr-developers/numcodecs.git
 
 Creating a development environment
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -75,7 +75,14 @@ The NumCodecs source code is hosted on GitHub at the following location:
 You will need your own fork to work on the code. Go to the link above and hit
 the "Fork" button. Then clone your fork to your local machine::
 
-    $ git clone --recursive git@github.com:your-user-name/numcodecs.git
+    $ git clone --recursive git@github.com:your-user-name/numcodecs.git  # with ``ssh``
+    
+    or
+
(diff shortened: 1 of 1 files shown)


docs/contributing.rst
```diff
@@ -75,9 +75,9 @@ The NumCodecs source code is hosted on GitHub at the following location:
 You will need your own fork to work on the code. Go to the link above and hit
 the "Fork" button. Then clone your fork to your local machine::
 
-    $ git clone git@github.com:your-user-name/numcodecs.git
```
joshmoore, line 78: This goal of this line (based on the proceeding text) is to checkout the user's own fork and not the main repo.

Saransh-cpp: Oh yes, missed that!

joshmoore: Thanks. So now my only question is if someone doesn't have SSH locally, will this allow them to push a change?

Saransh-cpp: Yes! I personally don't use SSH and it works fine.

jakirkham: Personally use `ssh` and know the `git@...` syntax is needed to push a change (`https://...` doesn't work)

Saransh-cpp: Updated!

jakirkham: Thanks! Though AIUI this doesn't work for you. So maybe we should show both syntaxes? Tried a pass at that below

Saransh-cpp:
Thanks for the suggestion! This renders like this right now -

image

jakirkham:
Thanks Saransh! 🙏

Let's see what Josh thinks 🙂

joshmoore:
> Let's see what Josh thinks

Sounds scary 👻 🕷️ ... ;) Anyway, LGTM. Auto-merged engaged! 🚀

docs/contributing.rst
```diff
@@ -75,9 +75,16 @@ The NumCodecs source code is hosted on GitHub at the following location:
...
+    
+or::
+
+    $ git clone --recursive https://github.com/your-user-name/numcodecs.git  # with ``https``
+
+Then ``cd`` into the clone and add the ``upstream`` remote::
+
     $ cd numcodecs
-    $ git remote add upstream git@github.com:zarr-developers/numcodecs.git
+    $ git remote add upstream git@github.com:zarr-developers/numcodecs.git  # or use ``https``
```
jakirkham, line 87: Actually would just use `https` here. It prevent pushing upstream when using `ssh`, which (to me at least) is a feature and not a bug :)

```suggestion
    $ git remote add upstream https://github.com/zarr-developers/numcodecs.git
```

jakirkham:
Thanks Josh!

Sorry had missed one thing. Added that above. Otherwise LGTM

PR #375, opened 2022-10-28
Drop headers

jakirkham:
Convert code from header files to Cython (generated C code is equivalent). Then drop the headers. These were there primarily for Windows compatibility with old Python versions (namely 2.7), but shouldn't be needed any more.

xref: https://github.com/zarr-developers/numcodecs/pull/368
xref: https://github.com/zarr-developers/numcodecs/pull/369

TODO:

- [ ] Unit tests and/or doctests in docstrings
- [ ] `tox -e py310` passes locally
- [ ] Docstrings and API docs for any new/modified user-facing classes and functions
- [ ] Changes documented in docs/release.rst
- [ ] `tox -e docs` passes locally
- [ ] GitHub Actions CI passes
- [ ] Test coverage to 100% (Coveralls passes)

PR #376, opened 2022-10-29
DEV: add Cython generated files to gitignore

tacaswell:
To make sure they do not accidentally get re-committed.

Dropped the checklist because this is not a code change and none of them were applicable.

Also 🎉 on removing the generated Cython output from source control!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -100,3 +100,6 @@ ENV/
 
 # setuptools-scm
 numcodecs/version.py
+
+# Cython generated
+numcodecs/*.c


tacaswell:
Oh, I see this is done as part of #375 , sorry for the noise.

jakirkham:
Thanks Tom! 🙏

No worries. Should be addressed now. Though if you spot anything else, please let us know 🙂

tacaswell:
I only noticed this because a `git merge` failed in https://github.com/tacaswell/build_the_world and I was poking around to see why.

Issue #379, opened 2022-11-02
`is_ndarray_like` has significant call overhead


ap--:
#### Problem description

Hello everyone,

I'm currently experimenting with decoding Aperio svs images via zarr directly to GPU. For that I tried to implement a [nvjpeg2k numcodecs-compatible](https://github.com/ap--/nvjpeg2k-numcodecs) decoder to decode directly to cupy ndarrays.

When profiling I noticed that a significant amount of time is spent on isinstance checks in `is_ndarray_like`. It seems that in the python typing module the Protocol meta-class generates the protocol attributes on every isinstance call instead of caching them.

In the profiling report below, you'll see that of the 77.3 seconds spent in `_chunk_getitem` 14.9 seconds are spent on `is_ndarray_like` (the bright pink block).

![Screenshot 2022-11-02 at 17 30 00](https://user-images.githubusercontent.com/1463443/199548689-e8240e89-46ee-44a4-8e7c-1972e64a2356.png)

I'll open a draft PR with a potential fix for discussion right away.

Cheers,
Andreas :smiley:

#### Version and installation information

Please provide the following:

* `numcodecs.__version__ == 0.10.2`
* Version of Python: 3.10.6
* Operating system: Linux
* Installed via conda

ap--:
Closed via #380 

In my benchmark total accumulated runtime of `is_ndarray_like` reduced from 14.9 to ~0.6 seconds.

jakirkham:
Thanks Andreas! 🎉

joshmoore:
🤯

PR #380, opened 2022-11-02
numcodecs.ndarray_like: speedup is_ndarray_like

ap--:
This PR reduces the runtime of `is_ndarray_like` and addresses #379

`is_ndarray_like()` calls `isinstance(obj, NDArrayLike)` where `NDArrayLike` is a runtime checkable Protocol class. For every check the protocol attributes are generated dynamically which can significantly contribute to the decode time in zarr arrays.

~This commit uses a pragmatic solution and duplicates the protocol attributes in the code base. This could be generated from `typing._get_protocol_attrs(NDArrayLike)` but would have to rely on private functionality of the typing module.~

TODO:

- [x] `tox -e py310` passes locally

### DIFF
--- a/numcodecs/ndarray_like.py
+++ b/numcodecs/ndarray_like.py
@@ -27,14 +27,14 @@ class _CachedProtocolMeta(Protocol.__class__):
 
 
 @runtime_checkable
-class DType(Protocol, metaclass=_CachedProtocolMeta):
+class DType(Protocol):
     itemsize: int
     name: str
     kind: str
 
 
 @runtime_checkable
-class FlagsObj(Protocol, metaclass=_CachedProtocolMeta):
+class FlagsObj(Protocol):
     c_contiguous: bool
     f_contiguous: bool
     owndata: bool
--- a/numcodecs/ndarray_like.py
+++ b/numcodecs/ndarray_like.py
@@ -1,5 +1,5 @@
 import sys
-from typing import Any, Optional, Tuple
+from typing import Any, Dict, Optional, Tuple, Type
 
 if sys.version_info >= (3, 8):
     from typing import Protocol, runtime_checkable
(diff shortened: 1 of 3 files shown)


numcodecs/ndarray_like.py
```diff
@@ -51,6 +51,26 @@ def view(self, dtype: DType = ...) -> "NDArrayLike":
         ...
 
 
+# keep in sync with NDArrayLike
+_NDArrayLike_protocol_attrs = [
```
martindurant, line 55: Nice and simple, but do we need all of these? I assume looping through takes some time too. I am surprised not to see the `__array*__` methods, though (NEP18 etc).

jakirkham:
@madsbk, could you please take a look when you have a chance?

numcodecs/ndarray_like.py
ap--: A subset would probably be enough. I can look through the zarr codebase tomorrow to check what's currently used. I just based it on the NDArrayLike class to be consistent. Currently total accumulated runtime of is_ndarray_like reduces from 14.9 to 2.8 seconds in my benchmark.

madsbk (commented):
Thanks @ap--, nice speedup!

I have two orthogonal suggestions:
- Move the check inside a `__instancecheck__`, which makes all use of `isinstance(obj, NDArrayLike)` faster not only `is_ndarray_like()`.
- Cache the result of `isinstance()` instead of checking the attributes manually. This way, we don't have to maintain an attribute list and we also check the type of `dtype` and `flags`.


It could look something like:
```python

from typing import Dict, Protocol, Type

class _ProtocolMeta(Protocol.__class__):  # type: ignore
    """Custom implementation of @runtime_checkable

    The native implementation of @runtime_checkable is slow,
    see <https://github.com/zarr-developers/numcodecs/issues/379>.

    Using caching, we speeds it up quite a bit.
    """

    _instancecheck_cache: Dict[Type, bool] = {}

    def __instancecheck__(cls, instance):
        ret = cls._instancecheck_cache.get(instance.__class__, None)
        if ret is None:
            ret = super().__instancecheck__(instance)
            cls._instancecheck_cache[instance.__class__] = ret
        return ret

@runtime_checkable
class NDArrayLike(Protocol, metaclass=_ProtocolMeta):
    dtype: DType
    shape: Tuple[int, ...]
    ...
```
We could consider limiting the maximum  number of cached values, but I don't think it matters. We are only saving a type and boolean per cache item and I cannot imagine this growing to much above 100 items.

ap--:
Sounds good 👍 
I'll implement the changes and update the PR.

ap--:
I updated the PR based on your suggested changes @madsbk .

After running my benchmark the cache dict looks like this:
```python
{
  <class 'bytes'>: False,
  <class 'numpy.ndarray'>: True,
  <class 'cupy.ndarray'>: True,
}
```
I think for all practical purposes this is unlikely to grow significantly more.

The total runtime of `is_ndarray_like` is now down to 0.58 seconds from 17.9 seconds before the PR 🎉 

Cheers,
Andreas

ap--:
So this is ready for review.
The lgtm error is a false positive. Should that be ignored via a comment in the code?

jakirkham:
Thanks Andreas! 🙏

As to the linter error, it appears according to [the Python docs `__instancecheck__` takes `self`]( https://docs.python.org/3/reference/datamodel.html?highlight=__instancecheck__#class.__instancecheck__ ). It is just `self` turns out to be a type (as opposed to this being a `classmethod`). So maybe that is a valid lint?

jakirkham:
LGTM. Only other suggestion would be to [mention this in the release notes]( https://github.com/zarr-developers/numcodecs/blob/main/docs/release.rst#unreleased )

ap--:
Updated the release notes :+1:

numcodecs/ndarray_like.py
jakirkham: This has been dropped after adopting Mads' suggestion

jakirkham (approved):
Thanks Andreas for the PR and Martin and Mads for the reviews! 🙏

Merging after CI passes

PR #387, opened 2022-11-04
Delete Python 2 leftover

DimitriPapadopoulos:
Detected by [pyugrade](https://github.com/asottile/pyupgrade).

TODO:

- [ ] Tests pass locally
- [ ] Docs build locally
- [X] GitHub Actions CI passes
- [X] Test coverage to 100% (Coveralls passes)

### DIFF
--- a/numcodecs/tests/common.py
+++ b/numcodecs/tests/common.py
@@ -276,7 +276,7 @@ def check_backwards_compatibility(codec_id, arrays, codecs, precision=None, pref
                 with open(codec_fn, mode='w') as cf:
                     _json.dump(codec.get_config(), cf, sort_keys=True, indent=4)
             # load config and compare with expectation
-            with open(codec_fn, mode='r') as cf:
+            with open(codec_fn) as cf:
                 config = _json.load(cf)
                 assert codec == get_codec(config)
 
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -34,7 +34,8 @@ Maintenance
   By :user:`John Kirkham <jakirkham>`, :issue:`375`.
 
 * Remove Python 2 code.
-  By :user:`Dimitri Papadopoulos Orfanos <DimitriPapadopoulos>` :issue:`368`.
+  By :user:`Dimitri Papadopoulos Orfanos <DimitriPapadopoulos>` :issue:`368`,
+  :issue:`387`.
 
 * Support Python 3.11.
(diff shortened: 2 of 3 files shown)


numcodecs/tests/common.py
```diff
@@ -276,7 +276,7 @@ def check_backwards_compatibility(codec_id, arrays, codecs, precision=None, pref
                 with open(codec_fn, mode='w') as cf:
                     _json.dump(codec.get_config(), cf, sort_keys=True, indent=4)
             # load config and compare with expectation
-            with open(codec_fn, mode='r') as cf:
+            with open(codec_fn) as cf:
```
jakirkham, line 279: Could we keep the `mode` here? Explicit is better than implicit

jakirkham (commented):
Thanks Dimitri! 🙏

Had one comment below

numcodecs/tests/common.py
DimitriPapadopoulos: I understand, but [pyupgrade](https://github.com/asottile/pyupgrade) or [refurb](https://github.com/dosisod/refurb) disagree and believe implicit is better than explicit :smile:

I believe the rationale is that [open()](https://docs.python.org/3/library/functions.html#open) has many default arguments, and all of them are implicit (_buffering_, _encoding_, _errors_, _newline_, _closefd_, _opener_). That's why Python has default arguments. Why make an exception for _mode_?

docs/conf.py
```diff
@@ -1,5 +1,4 @@
 #!/usr/bin/env python3
-# -*- coding: utf-8 -*-
```
DimitriPapadopoulos, line 2: Ooops. This is a generated file. Can't change it.

numcodecs/tests/common.py
DimitriPapadopoulos: But then I don't have strong opinions on the subject. I was myself rather reluctant initially:
https://github.com/codespell-project/codespell/pull/2364#discussion_r997742025

jakirkham (approved):

numcodecs/tests/common.py
jakirkham: Yeah `mode` seems special since it affects whether the file may be modified (or even truncated) when opening

That said, not too worried about it given the placement (in the tests)

Have occasionally run into unexpected default behavior around `mode` in some libraries. Hence the initial pause

PR #398, opened 2022-11-14
Use `conda-incubator/setup-miniconda@v2.2.0` (and use Conda on Linux)

jakirkham:
Update to the recent `conda-incubator/setup-miniconda` and pin that release. Also update Linux CI to also use Conda (just like on macOS & Windows). Drop Fortran compiler from CI as it is unused. Additionally drop `ACTIONS_ALLOW_UNSECURE_COMMANDS` as it is no longer needed. Some other small tidying.

xref: https://github.com/zarr-developers/zarr-python/pull/1263

<hr>

TODO:

- [ ] Tests pass locally
- [ ] Docs build locally

### DIFF
--- a/.github/workflows/ci-linux.yaml
+++ b/.github/workflows/ci-linux.yaml
@@ -17,22 +17,50 @@ jobs:
           submodules: recursive
 
       - name: Set up Python
-        uses: actions/setup-python@v4
+        uses: conda-incubator/setup-miniconda@v2.2.0
         with:
-          python-version: ${{ matrix.python-version }}
+            channels: conda-forge
+            python-version: ${{ matrix.python-version }}
+        env:
+            ACTIONS_ALLOW_UNSECURE_COMMANDS: true
+
+      - name: Show info about `base` environment
+        shell: "bash -l {0}"
+        run: |
+          conda info
+          conda config --show-sources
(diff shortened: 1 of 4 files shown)


joshmoore (approved):
Barring that this feels like a candidate for refactoring (_a la_ https://docs.github.com/en/actions/using-workflows/reusing-workflows) :+1: for unification.

.github/workflows/ci-linux.yaml
```diff
@@ -4,36 +4,67 @@ on: [push, pull_request]
...
+      - name: Show info about `env` environment
+        shell: "bash -l {0}"
+        run: |
+          conda list --show-channel-urls -n env
 
       - name: Install numcodecs
+        shell: "bash -l {0}"
         run: |
+          conda activate env
+          export DISABLE_NUMCODECS_AVX2=""
```
jakirkham, line 48: This was added since there were issues compiling with AVX2 on Linux. Would be good to revisit why this is occurring. That said, this is already [how we build in conda-forge]( https://github.com/conda-forge/numcodecs-feedstock/blob/ad779a7eaa6efcfe2ea54f8f0ba637d87ebaf1a3/recipe/meta.yaml#LL16 ).

PR #412, opened 2022-12-20
implement fletcher32

### LINKED ISSUE #410: Fletcher checksum codec
It would be great if numcodecs could implement the [Fletcher checksum](https://en.wikipedia.org/wiki/Fletcher%27s_checksum) algorithm in a way compatible with how hdf5 does it. In this version, the checksum is simply appended to the end of the chunk bytes. See https://github.com/fsspec/kerchunk/pull/274 for a dummy passthrough version.

There are some python implementations here:
- https://github.com/njaladan/hashpy/blob/master/hashpy/fletcherNbit.py
- https://gist.github.com/globby/9337839
- https://gist.github.com/AJPoulter-Soton/9d0d2505af64f0719bdee59b9a4533ba

martindurant:
Fixes #410 cc @rabernat 

TODO:

- [x] Tests pass locally
- [x] Docs build locally


martindurant:
The errors showing up are for non-numpy inputs to JSON and msgpack - nothing to do with this PR.

I'll fill in the docs and such shortly.

numcodecs/fletcher32.pyx
```diff
@@ -0,0 +1,54 @@
...
+    def decode(self, buf, out=None):
+        b = ensure_contiguous_ndarray(buf).view('uint8')
+        if len(buf) % 2:
+            # rare, odd size of bytes data only
+            arr = np.frombuffer(b.tobytes() + b"\x00", dtype="uint16")
+            val = fletcher32(arr)
+        else:
+            val = fletcher32(b[:-4].view('uint16'))
+        found = b[-4:].view('uint32')[0]
+        assert val == found
```
rabernat, line 50:
```suggestion
        if val != found:
            raise ValueError(
                f"The flecher32 checksum of the data ({found}) did not match the expected checksum ({val}). "
                "This could be a sign that the data has been corrupted."
            )
```

I think a verbose error would be much more useful to users than an assertion.

numcodecs/tests/test_fletcher32.py
```diff
@@ -0,0 +1,24 @@
...
+
+@pytest.mark.parametrize(
+    "dtype",
+    ["uint8", "int32", "float32"]
+)
+def test_with_data(dtype):
+    data = np.empty(100, dtype=dtype)
+    f = Fletcher32()
+    arr = np.frombuffer(f.decode(f.encode(data)), dtype=dtype)
+    assert (arr == data).all()
```
rabernat, line 24: In this module we should try mutating the data and verifying that the proper error is raised.

rabernat:
This looks amazing Martin! 🚀  Thanks so much for doing it.

One question: have your verified that this implementation is interoperable with the hdf5 and netcdf4 implementation? Like, if netcdf4 writes a chunk with fletcher32, does this codec successfully decode it?

martindurant:
> One question: have your verified that this implementation is interoperable with the hdf5 and netcdf4 implementation? Like, if netcdf4 writes a chunk with fletcher32, does this codec successfully decode it?

No, not yet. The tests just come from examples on wikipedia. Do you think we should bundle a small hdf file, or perhaps just extract a bytes buffer into a test?

rabernat:
> Do you think we should bundle a small hdf file, or perhaps just extract a bytes buffer into a test?

I would try to get a short snippet of actual bytes. I'd probably use the approach from https://github.com/fsspec/kerchunk/pull/274 of generating a tiny netcdf file using xarray with fletcher32 on, extracting a chunk manually using kerchunk, printing the bytes to the terminal, and then copy-pasting that to this PR.

Here is an example of such a string of bytes.

```python
b'x\xda\xb3\xf1\x8b\xd3gdb\x00\x02\xf1\xa3\x00'
```

martindurant:
Well it was a good idea to check! Their implementation is [not the same](https://github.com/Unidata/netcdf-c/blob/8eb71290eb9360dcfd4955ba94759ba8d02c40a9/plugins/H5checksum.c#L95), so I thought it best to just embed it directly. This ought to be faster too, if that's important.

numcodecs/fletcher32.pyx
```diff
@@ -0,0 +1,45 @@
...
+        return buf.tobytes() + struct.pack("<I", val)
+
+    def decode(self, buf, out=None):
+        """Check fletcher checksum, and return buffer without it"""
+        b = ensure_contiguous_ndarray(buf).view('uint8')
+        cdef const uint8_t[::1] b_ptr = b
+        val = H5_checksum_fletcher32(&b_ptr[0], b.nbytes - 4)
+        found = b[-4:].view("<u4")[0]
+        if val != found:
+            raise ValueError(
```
rabernat, line 37:
```suggestion
            raise RuntimeError(
```

numcodecs/tests/test_fletcher32.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def test_error():
+    data = np.arange(100)
+    f = Fletcher32()
+    enc = f.encode(data)
+    enc2 = bytearray(enc)
+    enc2[0] += 1
+    with pytest.raises(ValueError) as e:
```
rabernat, line 24:
```suggestion
    with pytest.raises(RuntimeError) as e:
```

numcodecs/tests/test_fletcher32.py
```diff
@@ -0,0 +1,42 @@
...
+        b'\xf2\xc6a\x01a\xb8\xe8?#&\x96\xabY\xf2\xe7?\xe2Pw\xba\xd0w\xea?'
+        b'\x80\xc5\xf8M@0\x9a?\x98H+\xb4\x03\xfa\xc6?\xb9P\x1e1'
+    )
+    data3 = Fletcher32().decode(data)
+    outarr = np.frombuffer(data3, dtype="<f8")
+    expected = [
+        0.0691225, 0.98130367, 0.87104532, 0.19018153, 0.9898866,
+        0.77250719, 0.74833377, 0.8271259, 0.02557469, 0.17950484
+    ]
+    assert np.allclose(outarr, expected)
```
rabernat, line 42: I feel that a test using integer data and an exact equality assertion would be more robust.

rabernat (commented):
Actually `RuntimeError` would be more consistent with what other codecs do when decompression fails.

numcodecs/_fletcher.c
```diff
@@ -0,0 +1,43 @@
+#include <stdint.h>
+#include <stddef.h>
+
+// https://github.com/Unidata/netcdf-c/blob/8eb71290eb9360dcfd4955ba94759ba8d02c40a9/plugins/H5checksum.c
+
+
+uint32_t H5_checksum_fletcher32(const void *_data, size_t _len)
```
jakirkham, line 7: Can we please move this to Cython? We've dropped all C code from this repo

martindurant: This is not generated C code, though, and cythonising something that already is C is both painful and inefficient

numcodecs/fletcher32.pyx
```diff
@@ -0,0 +1,45 @@
+import struct
+
+from numcodecs.abc import Codec
+from numcodecs.compat import ensure_contiguous_ndarray
+
+from libc.stdint cimport uint8_t, uint16_t, uint32_t
```
jakirkham, line 6: Does this need to be in Cython? It seems most of the functionality here can be done in Python. If the C file moves to Cython, this should be doable in pure Python. This should simplify things significantly from a maintenance standpoint. Things like inheriting from Python classes in Cython can sometimes be wonky

numcodecs/fletcher32.pyx
```diff
@@ -0,0 +1,45 @@
...
+    ValueError if inconsistent.
+    """
+
+    codec_id = "fletcher32"
+
+    def encode(self, buf):
+        """Return buffer plus 4-byte fletcher checksum"""
+        buf = ensure_contiguous_ndarray(buf).ravel().view('uint8')
+        cdef const uint8_t[::1] b_ptr = buf
+        val = H5_checksum_fletcher32(&b_ptr[0], buf.nbytes)
```
jakirkham, line 27: At the Cython level this could just be a function that takes a `const uint8_t[::1]` object, which would allow passing the buffer through unchanged without the need to access the address, size, etc.

numcodecs/fletcher32.pyx
```diff
@@ -0,0 +1,45 @@
...
+        return buf.tobytes() + struct.pack("<I", val)
+
+    def decode(self, buf, out=None):
+        """Check fletcher checksum, and return buffer without it"""
+        b = ensure_contiguous_ndarray(buf).view('uint8')
+        cdef const uint8_t[::1] b_ptr = b
+        val = H5_checksum_fletcher32(&b_ptr[0], b.nbytes - 4)
```
jakirkham, line 34: Same here

numcodecs/fletcher32.pyx
martindurant: I know that - it's the kind of thing I've done a bunch of times in the fastparquet codebase. However, there's  casting and prefix increments that are hard to follow in the embedded code. I just know that my earlier simpler cython version did not get the same result.

numcodecs/_fletcher.c
jakirkham: It's really not. For example [this header was in C]( https://github.com/zarr-developers/numcodecs/blob/cde8d624c5516854a5498df0be7fb1c7c5e766ef/numcodecs/stdint_compat.h ) and we moved it to [Cython]( https://github.com/zarr-developers/numcodecs/blob/955019b55a31b4a03c76f62edfde5df2301282c4/numcodecs/_utils.pxd ). This also stripped out a bunch of boilerplate that was no longer needed.

martindurant:
OK, @jakirkham , it didn't turn out to be too bad. The algorithm is obviously the same as the original.

I didn't understand why the class should be moved to a different pure-python module, though. lz4, blosc, zstd and vlen all have Codecs in their respective pyx files. (I must say, they are surprisingly complex!)

numcodecs/fletcher32.pyx
```diff
@@ -0,0 +1,79 @@
...
+    return (sum2 << 16) | sum1
+
+
+class Fletcher32(Codec):
+    """The fletcher checksum with 16-bit words and 32-bit output
+
+    With this codec, the checksum is concatenated on the end of the data
+    bytes when encoded. At decode time, the checksum is performed on
+    the data portion and compared with the four-byte checksum, raising
+    ValueError if inconsistent.
```
rabernat, line 52:
```suggestion
    RuntimeError if inconsistent.
```

jakirkham:
> OK, @jakirkham , it didn't turn out to be too bad. The algorithm is obviously the same as the original.

Thanks Martin! 🙏

> I didn't understand why the class should be moved to a different pure-python module, though. lz4, blosc, zstd and vlen all have Codecs in their respective pyx files. (I must say, they are surprisingly complex!)

Yeah we have some technical debt to work through for sure.

Since these are now in the same file, agree this matches the existing pattern.

Though at some point we might want to split these apart to simplify things. We need not do that here.

rabernat:
Do we have an existing issue to track the test failures we are seeing in this PR? As @martindurant says, they are not related to the new codec. But they will need to be fixed asap.

rabernat (approved):
This is a very useful contribution to numcodecs. Thanks a lot Martin! Provided we understand why the tests are failing and have a plan to fix that elsewhere, I'm happy to see this merged.

rabernat:
This has lingered for a while, but I think it looks good. We should get this in.

I have no idea what's going on the the tests. Maybe they are fixed by #417? Martin do you want to try to rebase?

joshmoore:
@martindurant @rabernat @jakirkham: did a release of this get discussed at any point while I was off galavanting?

martindurant:
No discussion I am aware of

jakirkham:
Raised an issue ( https://github.com/zarr-developers/numcodecs/issues/437 ) to discuss

PR #416, opened 2023-01-12
No ragged sequence for conversion to numpy arrays

bnavigator:
Numpy 1.24 does not accept ragged sequences anymore.

https://github.com/numpy/numpy/pull/22004


martindurant:
numpy should be able to support these as lists within an object array, no?
```
x = np.array([[1], [1, 2]], dtype="object")  # ok
```

bnavigator:
> numpy should be able to support these as lists within an object array, no?

That is correct, but the whole point of this roundtrip test ist that an inferred dtype which is not object is preserved.

Issue #420, opened 2023-02-21
libaec / szip support


mkitti:
NASA Earth Observing System (EOS) uses szip compression:
https://www.earthdata.nasa.gov/esdis/esco/standards-and-practices/hdf-eos5

It is an integrated compression codec in HDF5:
https://portal.hdfgroup.org/display/HDF5/Szip+Compression+in+HDF+Products

Deutschen Klimarechenzentrum (DKRZ) has a freely available implementation available here under a 2-Clause BSD License:
https://gitlab.dkrz.de/k202009/libaec
Github Mirror
https://github.com/MathisRosenhauer/libaec

Conda-forge has libaec binaries here:
https://anaconda.org/conda-forge/libaec/files

They are produced by the following feedstock:
https://github.com/conda-forge/libaec-feedstock

Based on the outstanding use of SZIP and the availability of the libaec implementation, I recommend that numcodecs support SZIP / AEC compression.

mkitti:
Also szip support via libaec was added to conda-forge's build of HDF5 in https://github.com/conda-forge/hdf5-feedstock/pull/179

martindurant:
I don't know about the merits of if sz versus any other compressor, but if there is data out there using it, then at least kerchunk would like numcodecs to be able to deal with it. So we need a little cython code to link with the available binary? We will probably not want numcodecs to explicitly depend on libaec.

martindurant:
Although I can see szlib.h is included with libaec via conda-forge, there are no instructions on how to call the function that seems to be the one we want, SZ_BufftoBuffDecompress .

mkitti:
Note that there are two shared libraries
1. `libaec.so`
2. `libszip.so`

I think you want the second one?

```
$ objdump -T libaec.so

libaec.so:     file format elf64-x86-64

DYNAMIC SYMBOL TABLE:
0000000000000000      DF *UND*	0000000000000000 (GLIBC_2.2.5) free
0000000000000000  w   D  *UND*	0000000000000000  Base        _ITM_deregisterTMCloneTable
0000000000000000      DF *UND*	0000000000000000 (GLIBC_2.2.5) memset
0000000000000000      DF *UND*	0000000000000000 (GLIBC_2.2.5) calloc
0000000000000000      DF *UND*	0000000000000000 (GLIBC_2.2.5) memcpy
0000000000000000  w   D  *UND*	0000000000000000  Base        __gmon_start__
[...]

PR #422, opened 2023-03-06
Add sz using ctypes

martindurant:
Fixes https://github.com/zarr-developers/numcodecs/issues/420

@mkitti , this hard crashes as is, but it should be an OK template to work from.

TODO:

- [ ] Tests pass locally
- [ ] Docs build locally


numcodecs/sz.py
```diff
@@ -0,0 +1,84 @@
...
+        self.pix_per_block = pix_per_block
+        self.bits_per_pixel = bits_per_pixel
+        self.pix_per_scanline = pix_per_scanline
+
+    def decode(self, buf, out=None):
+        check_sz()
+        buf = memoryview(buf)
+        param = Params(self.mask, self.bits_per_pixel, self.pix_per_block, self.pix_per_scanline)
+        lout = int.from_bytes(buf[:4], "little")
+        dest_len = ctypes.c_int()
```
mkitti, line 59:
```suggestion
        dest_len = ctypes.c_int(lout)
```

I think we want to provide the allocated size of the buffer as an input to `BuffertoBuffDecompress`.

martindurant: OK - the comments seems to suggest that this was to be set by the decompressor. 

Still crashes the example once applied.

numcodecs/sz.py
```diff
@@ -0,0 +1,84 @@
...
+        ctypes.POINTER(Params)
+    ]
+
+    BufftoBuffDecompress = libsz.SZ_BufftoBuffDecompress
+    BufftoBuffCompress = libsz.SZ_BufftoBuffCompress
+    BufftoBuffCompress.restype = ctypes.c_int
+    BufftoBuffCompress.argtypes = [
+        ctypes.c_void_p, ctypes.POINTER(ctypes.c_size_t), ctypes.c_void_p, ctypes.c_size_t,
+        ctypes.POINTER(Params)
+    ]
```
mkitti, line 36:
```suggestion
    BufftoBuffDecompress = libsz.SZ_BufftoBuffDecompress
    BufftoBuffDecompress.restype = ctypes.c_int
    BufftoBuffDecompress.argtypes = [
        ctypes.c_void_p, ctypes.POINTER(ctypes.c_size_t), ctypes.c_void_p, ctypes.c_size_t,
        ctypes.POINTER(Params)
    ]
```

martindurant: Hah, that was silly.
Still crashed, though, even after my fixed type

numcodecs/sz.py
```diff
@@ -0,0 +1,83 @@
...
+        lout = int.from_bytes(buf[:4], "little")
+        dest_len = ctypes.c_size_t(lout)
+        p_dest_len = ctypes.pointer(dest_len)
+        if out is None:
+            out = ctypes.create_string_buffer(lout)
+        buf2 = ctypes.c_void_p.from_buffer_copy(buf, 4)  # shouldn't need copy, but wants writable
+        ok = BufftoBuffDecompress(
+            out, p_dest_len, buf2, buf.nbytes - 4, param
+        )
+        assert dest_len == lout  # that we got the expected number of bytes
```
mkitti, line 66:
```suggestion
        assert dest_len.value == lout  # that we got the expected number of bytes
```

mkitti (changes requested):
After these two changes, this works for me.

numcodecs/sz.py
```diff
@@ -0,0 +1,83 @@
...
+    def decode(self, buf, out=None):
+        check_sz()
+        buf = memoryview(buf)
+        param = Params(self.mask, self.bits_per_pixel, self.pix_per_block, self.pix_per_scanline)
+        lout = int.from_bytes(buf[:4], "little")
+        dest_len = ctypes.c_size_t(lout)
+        p_dest_len = ctypes.pointer(dest_len)
+        if out is None:
+            out = ctypes.create_string_buffer(lout)
+        buf2 = ctypes.c_void_p.from_buffer_copy(buf, 4)  # shouldn't need copy, but wants writable
```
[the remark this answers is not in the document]
mkitti: This also works.

```suggestion
        buf2 = ctypes.c_char_p(buf.obj[4:])
```

mkitti:
I was looking through the libaec source code regarding the options. Only two options seem to be used practically.

https://gitlab.dkrz.de/k202009/libaec/-/blob/master/src/sz_compat.c#L53-54

```
    co[SZ_MSB_OPTION_MASK] = AEC_DATA_MSB;
    co[SZ_NN_OPTION_MASK] = AEC_DATA_PREPROCESS;
```

```c
#define SZ_MSB_OPTION_MASK 16
#define SZ_NN_OPTION_MASK 32
```

Thus the functional value classes for `mask_options` are

| option_mask | endian | preprocess | 
| --- | --- | --- |
| 0 | little endian | EC |
| 16 | big endian | EC |
| 32 | little endian | NN (do preprocessing) |
| 48 | big endian | NN (do preprocessing) |

Preprocessing is described below.
> `AEC_DATA_PREPROCESS`: preprocessing input will improve compression efficiency if data samples are correlated. It will only cost performance for no gain in efficiency if the data is already uncorrelated.

Besides the szip compatability, we may also want to consider an interface to AEC directly.

martindurant:
Personally, I my only use case here is to allow SZIP to be read via kerchunk. Do you suppose anyone would really use this to compress data? I would defer AEC compression and more detailed options until someone asks for them (but that someone might be you, @mkitti ).

jakirkham:
Thanks Martin! 🙏

Is there a reason this is using `ctypes` as opposed to Cython?

martindurant:
> Is there a reason this is using ctypes as opposed to Cython?

A couple of reasons
- no extra compile stub in our overgrown setup file
- only one source file instead of a .pyx and .py wrapper as recommended in my last PR
- no need to depend on libaec at build time

martindurant:
I am happy to accept all those suggestions. Is there a way to do it without clicking through them all?

jakirkham:
At least from my experience on the deployment side, have found `ctypes` more challenging to work with since it makes decisions about where to look for shared libraries, which are not always right and sometimes needs some nudging. Relying on the system library loader doesn't tend to have these issues.

Maybe a bigger question worth asking here given some of the concerns noted above, is whether this needs to be in Numcodecs or whether it could be a 3rd party library that plugins into Numcodecs (using the entry point mechanism)?

martindurant:
> whether this needs to be in Numcodecs or whether it could be a 3rd party library

kerchunk can happily host it, since it already has some codecs. I don't mind. That does imply that reading the NASA data via kerchunk is probably the only use it'll get.

mkitti:
> I am happy to accept all those suggestions. Is there a way to do it without clicking through them all?

The best I know is to use the "add suggestion to batch" button. Sorry.

https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/incorporating-feedback-in-your-pull-request

mkitti:
> That does imply that reading the NASA data via kerchunk is probably the only use it'll get.

SZIP/AEC is a [CCSDS](https://public.ccsds.org/default.aspx) recommended standard for loseless since 1997 and most recently published in the [August 2020 Blue Book](https://public.ccsds.org/Pubs/121x0b3.pdf). It is also [one of two builtin HDF5 compression filters](https://portal.hdfgroup.org/display/HDF5/Using+Compression+in+HDF5). The main blocker for wide adoption was that SZIP was under a non-free license.

With the release of a [BSD licensed libaec](https://gitlab.dkrz.de/k202009/libaec) from DKRZ (German Climate Computing Center) I think we might see increased adoption of this, especially since it is now being distributed via conda-forge. After https://github.com/conda-forge/hdf5-feedstock/pull/179, I think it's worthwhile to consider first class support for SZIP / AEC in numcodecs. A key impediment to the adoption of compression other than ZLIB-based deflate is the lack of universal availability of the compression codec.

We learned a lot about SZIP from this pull request despite not having much documentation to work with. Between this and the C demo earlier, https://github.com/zarr-developers/numcodecs/issues/420#issuecomment-1455187948, a Cython version might not be very difficult if that is preferred.

joshmoore:
> I would defer AEC compression and more detailed options until someone asks for them (but that someone might be you, @mkitti ).

To the extent that adding wouldn't constitute a breaking change later, no?

> The best I know is to use the "add suggestion to batch" button. Sorry.

Agreed. I could only see that under "Files" and not in this view. I applied them, @martindurant.

mkitti:
We made flake, unhappy. Let me see if I can fix that.

mkitti:
> To the extent that adding wouldn't constitute a breaking change later, no?

The direct AEC interface should be a distinct codec here. I do not know what advantage it may have over the SZIP interface.

numcodecs/sz.py
```diff
@@ -0,0 +1,130 @@
...
+SZ_OK = AEC_OK
+SZ_OUTBUFF_FULL = 2
+
+SZ_NO_ENCODER_ERROR = -1
+SZ_PARAM_ERROR = AEC_CONF_ERROR
+SZ_MEM_ERROR = AEC_MEM_ERROR
+
+SZ_MAX_PIXELS_PER_BLOCK = 32
+SZ_MAX_BLOCKS_PER_SCANLINE = 128
+SZ_MAX_PIXELS_PER_SCANLINE = SZ_MAX_PIXELS_PER_BLOCK * SZ_MAX_BLOCKS_PER_SCANLINE
```
mkitti, line 32:
```suggestion
SZ_MAX_PIXELS_PER_SCANLINE = SZ_MAX_PIXELS_PER_BLOCK * SZ_MAX_BLOCKS_PER_SCANLINE
```

numcodecs/sz.py
```diff
@@ -0,0 +1,130 @@
...
+
+SZ_NO_ENCODER_ERROR = -1
+SZ_PARAM_ERROR = AEC_CONF_ERROR
+SZ_MEM_ERROR = AEC_MEM_ERROR
+
+SZ_MAX_PIXELS_PER_BLOCK = 32
+SZ_MAX_BLOCKS_PER_SCANLINE = 128
+SZ_MAX_PIXELS_PER_SCANLINE = SZ_MAX_PIXELS_PER_BLOCK * SZ_MAX_BLOCKS_PER_SCANLINE 
+
+class Params(ctypes.Structure):
```
mkitti, line 34:
```suggestion

class Params(ctypes.Structure):
```

martindurant:
Can we agree, then, that this codec is useful enough to include here as-is?
A following aec codec could be considered in a separate PR.

I would argue that the ctypes implementation is actually simpler and better than cython for this case where are simply calling library functions.

jakirkham:
> Can we agree, then, that this codec is useful enough to include here as-is?

Usefulness is not really the discussion. Instead think we are discussing where it lives

In particular this becomes a conversation about expectations around ownership, maintenance, and what is considered required/optional

This already comes up with the various stores, their associated maintenance costs, and who owns them ( https://github.com/zarr-developers/zarr-python/issues/414 ) ( https://github.com/zarr-developers/zarr-python/issues/764 ) ( https://github.com/zarr-developers/zarr-python/issues/1274 ). The general consensus there has been we want to separate those out somehow

Think we have similar concern regarding compressors, which we have already collected a lot of. Some of which are domain specific

Given we already have an entrypoint system, we are in a much better place for making decisions about what batteries included means

In particular I'd push for this to be external. Think we should also do an audit of what is included here and decide whether we split it out.

>> whether this needs to be in Numcodecs or whether it could be a 3rd party library
> kerchunk can happily host it, since it already has some codecs. I don't mind. That does imply that reading the NASA data via kerchunk is probably the only use it'll get.

While these are 2 valid options, by no means are they the only ones.

For example, we could add a `numcodecs-szip` repo in this org and give relevant folks permissions. It could also live outside the org if preferred (have a slight preference for within the org, but not a strong one).

We can probably think of other choices.

> I would argue that the ctypes implementation is actually simpler and better than cython for this case where are simply calling library functions.

This is a secondary discussion from my perspective. It ties back into maintenance and ownership, but once those are settled then this may be moot.

mkitti:
If I were to name a distinct package for this code, I would name it numcodecs-aec.

Is this the right place to be having a conversation about how to split up numcodecs though? At most, we are adding an _optional_ binary dependency of less than 200 KB.

martindurant:
Whilst entrypoints is convenient for finding codecs installed in an environment, it is no help at all in dealing with missing codecs - we don't know how to instruct the user on what they need to install. Thus, there is a lot to be gained by including as many codecs as reasonable in the main repo or other standard collections (thinking of imagecodecs). Also, if we were to change the API, for instance to pass optional context information when en/decoding, having codecs live elsewhere pretty much ensures breaking workflows, at least temporarily.

Intake has the same problem, leading to https://intake.readthedocs.io/en/latest/plugin-directory.html and manual maintenance of links, which is also painful.

martindurant:
Closing in favour of the implementation in imagecodecs https://github.com/cgohlke/imagecodecs/blob/a74e4f305a9c8c317da17beb7a0698543c27b4ae/imagecodecs/numcodecs.py#L1408-L1450

Issue #342, opened 2022-07-19
arm64 wheel for Mac M1?


jonblower:
Hi, I recently came across an issue when installing numcodecs as a dependency of kerchunk on an M1 Mac (via `pip`). I was aiming to get the latest version (0.10.0) but ended up with 0.7.3, presumably because it was not able to build the latest version. I was then missing the "shuffle" codec. See https://github.com/fsspec/kerchunk/issues/195.

There is a 0.10.0 build in conda-forge that works. Is it possible to make this available via pip please?

PR #427, opened 2023-04-03
[CI] Use concurrency to cancel-in-progress workflows on new commit

psobolewskiPhD:
This PR adds concurrency and cancel-in-progress to the CI workflows. See:
https://docs.github.com/en/actions/using-jobs/using-concurrency
What this means is if workflows are running when a new commit is pushed, they are canceled and restarted, rather than piling up.

I tested this on my fork, see: https://github.com/psobolewskiPhD/numcodecs/actions/runs/4600373158

TODO:

- [ ] Unit tests and/or doctests in docstrings
- [ ] Tests pass locally
- [ ] Docstrings and API docs for any new/modified user-facing classes and functions
- [ ] Changes documented in docs/release.rst
- [ ] Docs build locally
- [ ] GitHub Actions CI passes
- [ ] Test coverage to 100% (Codecov passes)

PR #428, opened 2023-04-03
[CI] use CIBW overrides to build arm64 macOS wheels

psobolewskiPhD:
At present numcodecs doesn't have wheels for arm64 macOS.
It builds from source just fine, but not everyone may have development tools.
Wheels make things potentially safer and of course faster.

closes: https://github.com/zarr-developers/numcodecs/issues/342

In this PR I add use of CIBW_ARCHS_MACOS to the existing workflow, setting 'x86_64 arm64'. This results in wheels cross compiled for arm64 in addition to the x86 ones. The runners are x86, so the arm64 wheels won't be tested.
However, this also requires using `overrides` to ensure that both AVX2 and SSE2 are disabled.
As a result, the ENVIROMENT aspects are moved to pyproject.toml

I tested this on my fork: https://github.com/psobolewskiPhD/numcodecs/pull/2 
I dowloaded the arm64 wheels, pip installed locally, and ran pytest:
```
====== 583 passed, 33 skipped, 36 xfailed, 1 xpassed, 5 warnings in 8.84s ======
```
Another option is to use `universal2` and build a FAT wheel for x86 and arm64. The existing wheels arn't too big, so this isn't so bad? But I assume arm64 is still in the minority, so maybe separate wheels are better?

Also from a CI time PoV, it's worth considering whether full pytest needs to run on wheels on top of the CI tests for each PR/push, see: https://github.com/zarr-developers/numcodecs/pull/427#issuecomment-1494928538

BTW: this same approach may be a cleaner way to implement https://github.com/zarr-developers/numcodecs/pull/315

TODO:

- [x] Tests pass locally
- [ ] Docs build locally

### DIFF
--- a/.github/workflows/wheel.yaml
+++ b/.github/workflows/wheel.yaml
@@ -14,9 +14,10 @@ jobs:
       CIBW_TEST_COMMAND: pytest --pyargs numcodecs
       CIBW_TEST_REQUIRES: pytest
       CIBW_SKIP: "pp* cp36-* *-musllinux_* *win32 *_i686 *_s390x"
-      CIBW_ENVIRONMENT: "DISABLE_NUMCODECS_AVX2=1"
-      CIBW_ENVIRONMENT_MACOS: 'MACOSX_DEPLOYMENT_TARGET=10.9 DISABLE_NUMCODECS_AVX2=1 CFLAGS="$CFLAGS -Wno-implicit-function-declaration"'
-
+      CIBW_ARCHS_MACOS: 'x86_64 arm64'
+      CIBW_TEST_SKIP: '*-macosx_arm64'
+      # note: CIBW_ENVIRONMENT is now set in pyproject.toml
+      
     steps:
       - uses: actions/checkout@v3
         with:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -108,3 +108,11 @@ norecursedirs = [
     "notebooks",
     "numcodecs.egg-info",
 ]
+[tool.cibuildwheel]
+ environment = { DISABLE_NUMCODECS_AVX2=1 }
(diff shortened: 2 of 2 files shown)


joshmoore:
Thanks for this, @psobolewskiPhD. The primary worry to date has been on further extending the build time since we are approaching limits. Do you have a feeling for the overall increase here? (and/or options for keeping it under control)

psobolewskiPhD:
@joshmoore 
based on my fork, the macOS wheel step will take ~60% longer.
One strategy to reduce time would be to not duplicate testing, as I mention here:
https://github.com/zarr-developers/numcodecs/pull/427#issuecomment-1494928538

Another option is to not build wheels on PRs, but there are some advantages to that...

psobolewskiPhD:
Thinking about this specific PR, I assume one `universal2` wheel would be faster than building twice.

psobolewskiPhD:
Both PRs finished. Compared to the concurrency one (no arm64)
https://github.com/zarr-developers/numcodecs/actions/runs/4600452760/jobs/8150420382?pr=427
21.5 min
here 26 min, so not so bad?
Testing of the macOS ones is ~1 min per wheel...

psobolewskiPhD:
Gentle bump—I should have some time to return to this topic.

joshmoore:
Thanks again for the ping. I've updated the branch to check the current state and will go over the code now.

joshmoore:
@jakirkham: anything from your side here? I'm inclined to get it in (if still green) and then we let @psobolewskiPhD try to speed the builds up.

joshmoore:
I just rebuilt the failing job. Now all are green. I'm going to merge this to look at your gh-453. We can then start considering all the changes together for a release. Thanks again, @psobolewskiPhD.

PR #442, opened 2023-06-06
Remove use of entrypoints

s-t-e-v-e-n-k:
Since Python 3.8, the standard library has included functionality to query entry points directly using importlib.metadata. Since the API has changed for the better with Python 3.10, we need to support both ways of using it.

### DIFF
--- a/numcodecs/tests/test_entrypoints.py
+++ b/numcodecs/tests/test_entrypoints.py
@@ -7,7 +7,6 @@ import numcodecs.registry
 
 
 here = os.path.abspath(os.path.dirname(__file__))
-pytest.importorskip("entrypoints")
 
 
 @pytest.fixture()
@@ -20,7 +19,6 @@ def set_path():
     numcodecs.registry.codec_registry.pop("test")
 
 
-@pytest.mark.xfail(reason="FIXME: not working in wheels build")
 def test_entrypoint_codec(set_path):
     cls = numcodecs.registry.get_codec({"id": "test"})
     assert cls.codec_id == "test"
--- a/numcodecs/registry.py
+++ b/numcodecs/registry.py
@@ -1,7 +1,7 @@
 """The registry module provides some simple convenience functions to enable
 applications to dynamically register and look-up codec classes."""
+from importlib.metadata import entry_points
 import logging
-from contextlib import suppress
 
 logger = logging.getLogger("numcodecs")
 codec_registry = {}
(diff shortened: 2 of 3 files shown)


joshmoore:
(Triggered workflows)

joshmoore:
Makes sense to me. I've also opened (a currently failing) PR to remove entrypoint from the conda-forge package: https://github.com/conda-forge/numcodecs-feedstock/pull/95

jakirkham:
cc @martindurant

martindurant:
I don't know anything about `.select, but I trust you. We actually have tests exercising the entrypoint mechanism, right?

s-t-e-v-e-n-k:
There are tests, but they are marked as xfail.

joshmoore:
Good catch, @s-t-e-v-e-n-k. Can you un-xfail them here and we'll address whatever issues there are in the wheels build?

s-t-e-v-e-n-k:
I have rebased and removed the xfail.

joshmoore:
Re-launched workflows.

jakirkham:
Looks like the entrypoint test is failing with wheels

joshmoore:
```
      @pytest.fixture()
      def set_path():
          sys.path.append(here)
          numcodecs.registry.run_entrypoints()
          yield
          sys.path.remove(here)
          numcodecs.registry.run_entrypoints()
  >       numcodecs.registry.codec_registry.pop("test")
  E       KeyError: 'test'
```

This may be similar to the setup needed for conda-forge testing: i.e. it's not enough to just sys.path, but the test must actually be installed due to differing runtime environments.

s-t-e-v-e-n-k:
> This may be similar to the setup needed for conda-forge testing: i.e. it's not enough to just sys.path, but the test must actually be installed due to differing runtime environments.

I've been wondering about that, if we want to test this piece meal by mocking, and also doing an integration test by doing ... something like what the `sys.path` manipulation is trying to accomplish.

s-t-e-v-e-n-k:
I fell down an `importlib.metadata` vs `entrypoints` rabbit hole, but I nailed it down to the wheel not including the needed files for the tests to pass. I'm not certain why `MANIFEST.in` is not helping us here.

jakirkham:
RTD failures are due to an upstream issue ( https://github.com/readthedocs/readthedocs.org/issues/10601 ). Mentioned these instances there

jakirkham:
Looks like it may have been a GH issue. Restarted RTD and it appears to have passed. Also enabled GHA jobs to run

numcodecs/tests/test_entrypoints.py
```diff
@@ -20,7 +19,6 @@ def set_path():
     numcodecs.registry.codec_registry.pop("test")
 
 
-@pytest.mark.xfail(reason="FIXME: not working in wheels build")
```
joshmoore, line 23: Hurrah! :tada:

joshmoore (approved):
Thanks for all the hard work, @s-t-e-v-e-n-k! Merging lest other RTD issues crop up, but if there are any concerns from anyone, just let us know before the upcoming release.

PR #446, opened 2023-07-13
Add jenkins lookup3 32-bit checksum

### LINKED ISSUE #445: Add Jenkin's lookup3 as a 32-bit checksum for HDF5

```python
# Jenkin's lookup3 can be useful for verifying HDF5 data structures such as the superblock below
import jenkins_cffi

with open("original_hdf5_zarr_shard_demo.h5", "rb") as f:
    b = f.read(48)

hash_bytes = jenkins_cffi.hashlittle(bytes(b[:-4])).to_bytes(4, "little")
print(b[-4:] == hash_bytes) # True

```

#### Problem description

Jenkin's lookup3 is an integral component of the HDF5 specification for its internal datastructures.

This becomes relevant if we would like to reuse HDF5 data structures. For example, the HDF5 Fixed Array Data Block can made byte compatible with the proposed Zarr shard specification, except for the four byte checksum. Currently, the only permitted checksum is crc32.

https://github.com/zarr-developers/zarr-specs/pull/152#issuecomment-1632034874

An implementation of Bob Jenkin's lookup3 is widely available across many languages.

* https://en.wikipedia.org/wiki/Jenkins_hash_function
* https://www.burtleburtle.net/bob/hash/doobs.html
* C: https://www.burtleburtle.net/bob/c/lookup3.c
* C: https://github.com/HDFGroup/hdf5/blob/3af8bb267d6ad2d4e8e5d77e16d5a8f7625ad34d/src/H5checksum.c#L364-L458
* Python: https://pypi.org/project/jenkins/ (old PyPi package)
* Python: https://pypi.org/project/jenkins-cffi/
* JavaScript: https://www.npmjs.com/package/jenkins-hash-lookup3
* Julia: https://github.com/JuliaIO/JLD2.jl/blob/master/src/Lookup3.jl

Please provide the following:

```
jenkins-cffi              1.0.2.1                  pypi_0    pypi
python                    3.11.0          he550d4f_1_cpython    conda-forge
```

mkitti:
Add Bob Jenkin's lookup3 from HDF5 via a Cython implementation

Fix #445

TODO:

- [x] Tests pass locally
- [x] Docs build locally

### DIFF
--- a/numcodecs/jenkins.pyx
+++ b/numcodecs/jenkins.pyx
@@ -84,15 +84,13 @@ computer software, distribute, and sublicense such enhancements or derivative
 works thereof, in binary and source code form.
 """
 
-import struct
 import cython
 
-from numcodecs.checksum32 import Checksum32
 from numcodecs.compat import ensure_contiguous_ndarray
 
 from libc.stdint cimport uint8_t, uint16_t, uint32_t
 
-def jenkins_lookup3(data: bytes, initval: uint32_t=0):
+cpdef uint32_t jenkins_lookup3(const uint8_t[::1] _data, uint32_t initval=0):
     """
     jenkins_lookup3(data: bytes, initval: uint32_t=0)
     hash a variable-length key into a 32-bit value
@@ -114,7 +112,7 @@ def jenkins_lookup3(data: bytes, initval: uint32_t=0):
     ```
     h = 0
     for k in strings:
-      h = _jenkins_lookup3(k, h)
+      h = jenkins_lookup3(k, h)
     ```
 
     By Bob Jenkins, 2006.  bob_jenkins@burtleburtle.net.  You may use this
@@ -122,12 +120,6 @@ def jenkins_lookup3(data: bytes, initval: uint32_t=0):
 
     Use for hash table lookup, or anything where one collision in 2^^32 is
     acceptable.  Do NOT use for cryptographic purposes.
-    """
-    return _jenkins_lookup3(data, initval)
-
-cdef uint32_t _jenkins_lookup3(const uint8_t[::1] _data, uint32_t initval=0):
-    """
-    Implementation of jenkins_lookup3
 
     Converted from H5_checksum_lookup3
     https://github.com/HDFGroup/hdf5/blob/577c192518598c7e2945683655feffcdbdf5a91b/src/H5checksum.c#L378-L472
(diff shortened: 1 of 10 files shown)


mkitti:
@rabernat Here's a cython implementation of jenkin's lookup3 32-bit checksum.

mkitti:
I'm thinking of moving `initval` into the codec so that it is defined at the constructor.

Also, I'm thinking about a `prepend` value in case we need to add extra bytes to the beginning of the bytes being checksummed.

numcodecs/jenkins.pyx
```diff
@@ -0,0 +1,374 @@
...
+    With this codec, the checksum is concatenated on the end of the data
+    bytes when encoded. At decode time, the checksum is performed on
+    the data portion and compared with the four-byte checksum, raising
+    RuntimeError if inconsistent.
+    """
+
+    checksum = jenkins_lookup3
+    codec_id = "bob_jenkins_lookup3"
+
+    def encode(self, buf, initval=0):
```
martindurant, line 350: Indeed, initval will never be non-default like this - if you want to use it at all, it should be in `__init__` and stored as an attribute

numcodecs/jenkins.pyx
```diff
@@ -0,0 +1,374 @@
...
+
+import struct
+import cython
+
+from numcodecs.checksum32 import Checksum32
+from numcodecs.compat import ensure_contiguous_ndarray
+
+from libc.stdint cimport uint8_t, uint16_t, uint32_t
+
+def jenkins_lookup3(data: bytes, initval: uint32_t=0):
```
martindurant, line 95: This python version is not called anywhere?

numcodecs/jenkins.pyx
```diff
@@ -0,0 +1,374 @@
...
+
+    By Bob Jenkins, 2006.  bob_jenkins@burtleburtle.net.  You may use this
+    code any way you wish, private, educational, or commercial.  It's free.
+
+    Use for hash table lookup, or anything where one collision in 2^^32 is
+    acceptable.  Do NOT use for cryptographic purposes.
+    """
+    return _jenkins_lookup3(data, initval)
+
+cdef uint32_t _jenkins_lookup3(const uint8_t[::1] _data, uint32_t initval=0):
```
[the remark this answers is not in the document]
mkitti: Resolved in [296e4ea](https://github.com/zarr-developers/numcodecs/pull/446/commits/296e4eaf809a9d73c0ee3b295a99d64ea023cb79)

numcodecs/jenkins.pyx
mkitti: Resolved in [296e4ea](https://github.com/zarr-developers/numcodecs/pull/446/commits/296e4eaf809a9d73c0ee3b295a99d64ea023cb79)

numcodecs/jenkins.pyx
mkitti: cpdef is now used and this is called from the JenkinsLookup3 class

Resolved in [296e4ea](https://github.com/zarr-developers/numcodecs/pull/446/commits/296e4eaf809a9d73c0ee3b295a99d64ea023cb79)

numcodecs/jenkins.pyx
```diff
@@ -0,0 +1,374 @@
...
+    RuntimeError if inconsistent.
+    """
+
+    checksum = jenkins_lookup3
+    codec_id = "bob_jenkins_lookup3"
+
+    def encode(self, buf, initval=0):
+        """Return buffer plus 4-byte Bob Jenkin's lookup3 checksum"""
+        buf = ensure_contiguous_ndarray(buf).ravel().view('uint8')
+        cdef const uint8_t[::1] b_ptr = buf
```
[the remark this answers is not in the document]
mkitti: I moved the class into checksum32.py

Resolved in [296e4ea](https://github.com/zarr-developers/numcodecs/pull/446/commits/296e4eaf809a9d73c0ee3b295a99d64ea023cb79)

mkitti:
Could someone run CI?

martindurant:
Missing coverage: https://app.codecov.io/gh/zarr-developers/numcodecs/pull/446/blob/numcodecs/checksum32.py

numcodecs/jenkins.pyx
```diff
@@ -0,0 +1,327 @@
...
+        a, b, c = _jenkins_lookup3_mix(a, b, c)
+        length -= 12
+        k += 12
+
+    # -------------------------------- last block: affect all 32 bits of (c)
+    if length == 12:
+        c += (<uint32_t>k[11]) << 24
+        length -= 1
+        
+    if length == 11:
```
jakirkham, line 172: Should these be `elif`?

It's worth noting that Cython can detect cases like these when they are written as `if...elif...else` and convert them to `switch...case` in C/C++

mkitti:
Flake8 issues resolved

mkitti:
I made some changes to `docs/Makefile`. When I ran docs/Makefile without having sphinx on my path, I got the following cryptic Makefile error:
```
$ ~/src/numcodecs/docs$ make
Makefile:12: *** recipe commences before first target.  Stop.
```

This was due to the if statement near the top of the file detecting the lack of sphinx, and then proceeding to try to print the "friendly" error message. However, it accomplishes exactly the opposite.

mkitti:
I modified fletcher32 since the `out` keyword was not working in my hands. I also added a test for it.

mkitti:
Let me know if you vwould prefer me to break this up a bit. I originally intended that but slipped in a `commit -a` by accident.

martindurant:
Yes, please keep only the jenkins stuff in here. The other bits look OK I think, but it's better to keep PRs focussed where possible.

mkitti:
I split off unrelated changes into #449 (Fletcher32) and #451 (sphinx Makefile)

PR #449, opened 2023-07-18
Fix Fletcher32.decode out keyword and test

### LINKED ISSUE #448: Fletcher32 decode out keyword does not work

```python
In [1]: import numcodecs

In [2]: f = numcodecs.Fletcher32()

In [3]: result = f.encode( b"Hello world")

In [5]: import numpy as np

In [6]: f.decode(result, out = np.zeros(11, dtype="u1"))
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[6], line 1
----> 1 f.decode(result, out = np.zeros(11, dtype="u1"))

File ~/src/numcodecs/numcodecs/fletcher32.pyx:82, in numcodecs.fletcher32.Fletcher32.decode()
     80         "This could be a sign that the data has been corrupted."
     81     )
---> 82 if out:
     83     out.view("uint8")[:] = b[:-4]
     84     return out

ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()


```

#### Problem description

`Fletcher32.decode` contains an `out` keyword that seems to be intended for a `numpy.ndarray` due to the `.view` property. However, when provided with `numpy.ndarray` a value error occurs as per above.

Please provide the following:

* Python 3.11.4
* main git branch of numcodecs
* Linux

mkitti:
Fix #448 , enable `out` keyword of `Fletcher32.decode`.

- [x] Tests pass locally
- [x] Docs build locally

### DIFF
--- a/numcodecs/fletcher32.pyx
+++ b/numcodecs/fletcher32.pyx
@@ -79,7 +79,7 @@ class Fletcher32(Codec):
                 f" match the expected checksum ({found}).\n"
                 "This could be a sign that the data has been corrupted."
             )
-        if out:
+        if out is not None:
             out.view("uint8")[:] = b[:-4]
             return out
         return memoryview(b[:-4])
--- a/numcodecs/tests/test_fletcher32.py
+++ b/numcodecs/tests/test_fletcher32.py
@@ -40,3 +40,10 @@ def test_known():
         1911, -2427, 1897, -2412, 2440, 873, -621, -829, 551, -2118,
     ]
     assert outarr.tolist() == expected
+
+
+def test_out():
+    data = np.frombuffer(bytearray(b"Hello World"), dtype="uint8")
+    f = Fletcher32()
+    result = f.encode(data)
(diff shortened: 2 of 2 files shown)


mkitti:
Not sure if the conditional needs to be clarified, but the keyword was not tested. I realized this from looking at coverage of jenkins lookup3 checksum.

mkitti:
I am unclear if release notes are needed since Fletcher32 has not been released

mkitti:
@martindurant was this `out` keyword actually supposed to be used? The other thought I had was perhaps this is primarily vestigal and could just be removed.

martindurant:
Yes, I think `out` should be supported. I think zarr provides it when reading a whole chunk (rather than a slice) AND there are no further stages of decompression/decoding. The latter condition probably doesn't happen regularly in HDF5, which is where we find fletcher32, but it is possible.

mkitti:
Let me know if you think this is the correct fix or not then.

PR #463, opened 2023-09-13
Preliminary version of Blosc2 module

FrancescAlted:
This is a *preliminary* version of a `blosc2.pyx` module for numcodecs.  It is essentially the same than `blosc.pyx`, but the API has been updated to use Blosc2 instead.  A new C-Blosc2 submodule has been added too.

However, C-Blosc2 requires cmake to build, and we normally use something like [scikit-build-core](https://github.com/scikit-build/scikit-build-core). Unfortunately this means quite a bit of changes into the current build procedure based on setuptools; would you be up to adopt these changes?  If you don't, probably Blosc2 can be made to work with existing setuptools, but we still need to figure out how.

TODO:

- [ ] Update setup.py / pyproject.toml to compile the C-Blosc2 library
- [ ] Tests pass locally
- [ ] Docs build locally

### DIFF
--- a/numcodecs/blosc2.pyx
+++ b/numcodecs/blosc2.pyx
@@ -113,6 +113,9 @@ cdef extern from "blosc2.h":
 
     void blosc2_init()
     void blosc2_destroy()
+    blosc2_context* blosc2_create_cctx(blosc2_cparams cparams)
+    blosc2_context * blosc2_create_dctx(blosc2_dparams dparams)
+    void blosc2_free_ctx(blosc2_context * context)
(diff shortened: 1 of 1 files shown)


mkitti:
I think this is an important step to take in the right direction.

FrancescAlted:
> @FrancescAlted, do you have a timeline for how much longer c-blosc1 will be maintained?

Well, the Blosc team would be supporting C-Blosc1 for as long as possible.  Having said that, all the action is happening in C-Blosc2 right now, and C-Blosc1 will only receive fixes in the future.

PR #475, opened 2023-10-09
Fix default entrypoint handling on older Pythons

### LINKED ISSUE #478: Default entry point is being updated incorrectly

I don't think it's worth trying to create an MRE for this situation, please see below.

#### Problem description

This has turned out to be a much more subtle issue than I first thought. The problem looks like it's due to a combination of two factors:
1. The interface of the `EntryPoint` object changing over time, and 
2. The monkey-patching of `importlib.metadata` by `importlib_metadata`

To start with, we need to look at the evolution of `importlib_metadata`. Until `importlib_metadata` version 4.13.0, [the `EntryPoints` class had an `__iter__` method](https://github.com/python/importlib_metadata/blob/v4.13.0/importlib_metadata/__init__.py#L230). This method was deprecated much earlier (sometime around 3.x) but was not removed until version 5.0.0. Meanwhile, version 4.8.1 [introduced support for `EntryPoints.__getitem__` via the `DeprecatedTuple` class](https://github.com/python/importlib_metadata/blob/v4.8.1/importlib_metadata/__init__.py#L149), functionality which had been supported in previous versions but had been removed. 

Now, when we roll around to version 5.0.0, the removal of `EntryPoints.__iter__` meant that iteration over an `EntryPoints` object instead fell back to calling `__getitem__` ([this is how iterables are designed in Python](https://docs.python.org/3/glossary.html#term-iterable)). The problem is that the `__getitem__` implementation would in turn call `_key`, which would return a tuple `(self.name, self.value, self.group)`. This is a problem because the old `__iter__` implementation instead returned `iter((self.name, self))`.

[The relevant code in numcodecs](https://github.com/zarr-developers/numcodecs/blob/main/numcodecs/registry.py#L19) for Python 3.9 is assuming that we'll get the second interface, such that given a tuple of `EntryPoint` objects we'll be able to process it as key-value pairs of the form `(name, EntryPoint)`, which works with the old explicit `__iter__` API but not the new `__getitem__` implicit iteration API.

A major part of the complexity comes from the fact that `importlib_metadata` was provisionally added to the standard library as `importlib.metadata` in Python 3.8, then made a full part of the standard in 3.10. As of Python 3.9, the version of the code still had an `EntryPoints` object that defined `__iter__`, i.e. it was matching `importlib_metadata`<5.0.0. Therefore, the code in numcodecs seems like it should work. However, `importlib_metadata` actually monkey-patches parts of importlib.metadata:
```
>>> from importlib.metadata import entry_points
>>> type(entry_points()[next(iter(entry_points()))][0])
<class 'importlib.metadata.EntryPoint'>
>>> import importlib_metadata
>>> type(entry_points()[next(iter(entry_points()))][0])
<class 'importlib_metadata.EntryPoint'>
```
Note that on the final line we go from `importlib.metadata` to `importlib_metadata`. This is the root of the problem: if you have a newer version of `importlib_metadata` installed in Python 3.9, you start seeing `EntryPoints` objects with the new iteration protocol instead of the old one. Therefore, in the Python 3.9 scenario numcodecs needs to handle both possibilities.

The simplest solution is to avoid relying on the iteration protocol altogether and simply construct the appropriate update argument manually. This is what I now do in #475.

Please provide the following:

* Value of ``numcodecs.__version__``: 0.12.0
* Version of Python interpreter: 3.9.18
* Operating system (Linux/Windows/Mac): Linux (Ubuntu 20.04
* How NumCodecs was installed (e.g., "using pip into virtual environment", or "using conda"): Using conda or pip

Also, if you think it might be relevant, please provide the output from ``pip list`` or
``conda list`` depending on which was used to install NumCodecs.

vyasr:
The fallback here for updating a dictionary needs to be a dictionary, not a list.

I haven't added any new tests or anything, but I can if I'm pointed in the right direction. Not sure how this library wants to test these types of edge cases with entry points.

Resolves #478 

TODO:

- [x] Tests pass locally
- [ ] Docs build locally

### DIFF
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -7,6 +7,17 @@ Release notes
     # re-indented so that it does not show up in the notes.
 
 
+.. _release_0.12.1:
+
+0.12.1
+------
+
+Fix
+~~~
+
+* Fix default entrypoint handling on older Pythons
+  By :user:`Vyas Ramasubramani <vyasr>`, :issue:`475`.
+
 .. _release_0.12.0:
 
 0.12.0
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -16,7 +16,7 @@ Fix
 ~~~
 
 * Fix default entrypoint handling on older Pythons
-  By :user:`Vyas Ramasubramani <vyasr>`, :issue:`475`.
+  By :user:`Vyas Ramasubramani <vyasr>`, :issue:`478`.
 
 .. _release_0.12.0:
 
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -16,15 +16,6 @@ Fix
 
 * `Codec` is now derived from `abc.ABC`
   By :user:`Mads R. B. Kristensen <madsbk>`, :issue:`472`.
-
-.. _release_0.12.1:
-
-0.12.1
-------
-
-Fix
-~~~
-
 * Fix handling of entry points on older Python versions where importlib_metadata compatibility is concerned
   By :user:`Vyas Ramasubramani <vyasr>`, :issue:`478`.
 
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -15,7 +15,7 @@ Release notes
 Fix
 ~~~
 
-* Fix default entrypoint handling on older Pythons
+* Fix handling of entry points on older Python versions where importlib_metadata compatibility is concerned
   By :user:`Vyas Ramasubramani <vyasr>`, :issue:`478`.
 
 .. _release_0.12.0:
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -16,7 +16,7 @@ Fix
 
 * ``Codec`` is now derived from ``abc.ABC``
   By :user:`Mads R. B. Kristensen <madsbk>`, :issue:`472`.
-* Fix handling of entry points on older Python versions where importlib_metadata compatibility is concerned
+* Fix handling of entry points on older Python versions where ``importlib_metadata`` compatibility is concerned
   By :user:`Vyas Ramasubramani <vyasr>`, :issue:`478`.
 * Make shuffle pyx functions ``noexcept``
   By :user:`Martin Durant <martindurant>`, :issue:`477`.
--- a/numcodecs/registry.py
+++ b/numcodecs/registry.py
@@ -16,7 +16,7 @@ def run_entrypoints():
         entries.update({e.name: e for e in eps.select(group="numcodecs.codecs")})
     else:
         # Otherwise, fallback to using get
-        entries.update(eps.get("numcodecs.codecs", []))
+        entries.update(eps.get("numcodecs.codecs", {}))
 
 
 run_entrypoints()
--- a/numcodecs/registry.py
+++ b/numcodecs/registry.py
@@ -16,7 +16,7 @@ def run_entrypoints():
         entries.update({e.name: e for e in eps.select(group="numcodecs.codecs")})
     else:
         # Otherwise, fallback to using get
-        entries.update(eps.get("numcodecs.codecs", {}))
+        entries.update({e.name: e for e in eps.get("numcodecs.codecs", [])})
 
 
 run_entrypoints()
--- a/numcodecs/tests/test_entrypoints.py
+++ b/numcodecs/tests/test_entrypoints.py
@@ -4,6 +4,8 @@ from unittest import mock
 
 import pytest
 
+from multiprocessing import Process
+
 import numcodecs.registry
 
 
@@ -30,16 +32,19 @@ def test_entrypoint_codec():
     assert cls.codec_id == "test"
 
 
-def test_entrypoint_codec_with_importlib_metadata():
+def get_entrypoints_with_importlib_metadata_loaded():
     # importlib_metadata patches importlib.metadata, which can lead to breaking changes
     # to the APIs of EntryPoint objects used when registering entrypoints. Attempt to
     # isolate those changes to just this test.
-    with mock.patch.dict(sys.modules):
-        import importlib_metadata  # noqa: F401
-        sys.path.append(here)
-        numcodecs.registry.run_entrypoints()
-        cls = numcodecs.registry.get_codec({"id": "test"})
-        assert cls.codec_id == "test"
-        sys.path.remove(here)
-        numcodecs.registry.run_entrypoints()
-        numcodecs.registry.codec_registry.pop("test")
+    import importlib_metadata  # noqa: F401
+    sys.path.append(here)
+    numcodecs.registry.run_entrypoints()
+    cls = numcodecs.registry.get_codec({"id": "test"})
+    assert cls.codec_id == "test"
+
+
+def test_entrypoint_codec_with_importlib_metadata():
+    p = Process(target=get_entrypoints_with_importlib_metadata_loaded)
+    p.start()
+    p.join()
+    assert p.exitcode == 0
--- a/numcodecs/tests/test_entrypoints.py
+++ b/numcodecs/tests/test_entrypoints.py
@@ -3,8 +3,6 @@ import sys
 
 import pytest
 
-from multiprocessing import Process
-
 import numcodecs.registry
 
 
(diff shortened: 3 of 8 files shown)


jakirkham:
Thanks Vyas! 🙏

Made some tweaks to handle the release note. Hope that is ok. Please update as needed

Also maybe we should test this somehow. Currently we test the presence of a single entrypoint. Maybe we should also test the absence of any entrypoints?

https://github.com/zarr-developers/numcodecs/blob/a5539b5876fb76f014ac74c17fa98d805d5bbfda/numcodecs/tests/test_entrypoints.py#L22-L24

numcodecs/registry.py
```diff
@@ -16,7 +16,7 @@ def run_entrypoints():
         entries.update({e.name: e for e in eps.select(group="numcodecs.codecs")})
     else:
         # Otherwise, fallback to using get
-        entries.update(eps.get("numcodecs.codecs", []))
+        entries.update(eps.get("numcodecs.codecs", {}))
```
jakirkham, line 19: Looks like this particular line traces back to PR: https://github.com/zarr-developers/numcodecs/pull/442

cc @s-t-e-v-e-n-k (in case you have thoughts on this change)

s-t-e-v-e-n-k: I guess so? I will admit I was thinking about what `entry_points().get()` would return, rather than updating the dictionary. I would feel more comfortable if there was a test case with no entry points, but I apologise for not adding a test case myself.

vyasr: I'm trying to add a test now and I'm no longer certain I've isolated quite the right case that is triggering the issue. Stay tuned.

vyasr:
OK this fix isn't right. The issue isn't the fallback case, but rather a real case where numcodecs.codec is being registered as `None`. Looking into that now.

numcodecs/registry.py
vyasr: OK, I've written up my full post-portem in #478 now. The problem is far more esoteric than I initially diagnosed. I'm not sure a test here is going to be very feasible, please have a look at my writeup in the issue and let me know what you think.

jakirkham:
@martindurant , could you please look over this?

numcodecs/registry.py
vyasr: All right, I think I've found an OK way to write a test. Let's see if it really isolates well enough to avoid mucking with other tests though.

s-t-e-v-e-n-k: I will admit I'm not happy with how importlib_metadata plays with importlib.metadata, and that I didn't consider API breaking changes between them, since I was only testing with importlib.metadata, but other projects might utilise numcodecs and something else that requires importlib_metadata. :-(

vyasr: Yeah this is a pretty gross situation to wind up in. I haven't had to dive this deep in before and it took me a couple hours to really zero in on what was going on. Totally understandable how this could get missed in numcodecs. It feels like a problem that isn't really going to get fixed until Python 3.9 EOLs :disappointed:

numcodecs/tests/test_entrypoints.py
```diff
@@ -19,6 +19,27 @@ def set_path():
...
+def test_entrypoint_codec():
     cls = numcodecs.registry.get_codec({"id": "test"})
     assert cls.codec_id == "test"
+
+
+def test_entrypoint_codec_with_importlib_metadata():
+    # importlib_metadata patches importlib.metadata, which can lead to breaking changes
+    # to the APIs of EntryPoint objects used when registering entrypoints. Attempt to
+    # isolate those changes to just this test.
+    with mock.patch.dict(sys.modules):
```
vyasr, line 37: In hindsight, this may not really be enough to avoid this import polluting the rest of the test suite. Even if `importlib_metadata` is removed from `sys.modules`, the monkey-patching is already done. We might need to launch this test in a subprocess to really avoid this problem...

vyasr: OK I have a version of this test locally that's rewritten to use `multiprocessing.Process` and will fully isolate the import. I can push that version if maintainers are OK with a subprocess in the test.

vyasr: All right, since there were some unrelated test failures I pushed out the multiprocessing version anyway so that you can take a look. I can revert if it's preferred to avoid that.

jakirkham:
Was seeing AVX2 issues on macOS builds, that I thought we just fixed with PR ( https://github.com/zarr-developers/numcodecs/pull/479 ). So toggling PR to ensure latest changes in `main` are picked up

joshmoore:
OSX failures are the sporadic:
```
#error AVX2 is not supported by the target architecture/platform and/or this compiler.
```
Trying a rebuild.

vyasr:
Maybe one more rerun? Looks like attempt 2 passed one of the previously failing jobs but the other failed again.

jakirkham:
Have added a fix in `main` to disable AVX2 in macOS GHA jobs and merged that into here. That should resolve that issue

jakirkham:
Since the only entrypoint added is coming from the test entrypoint that we load ourselves. Wonder if we can avoid using a subprocess by simply testing there are no entrypoints present (and effectively those code paths work) before trying to add our own entrypoint (to test that we can load entrypoints). What do you think Vyas?

vyasr:
> Since the only entrypoint added is coming from the test entrypoint that we load ourselves. Wonder if we can avoid using a subprocess by simply testing there are no entrypoints present (and effectively those code paths work) before trying to add our own entrypoint (to test that we can load entrypoints). What do you think Vyas?

I'm afraid I don't understand how your suggestion avoids the problem. My concern is that because `importlib_metadata` is patching the standard library, as soon as the test importing it runs all subsequent tests in the same process will see the patched library. If there are any other edge cases where simply importing `importlib_metadata` causes problems, we would start seeing random unexpected failures in other tests based on test ordering (and of course partitioning if tests are run with `pytest-xdist`). At present, that mostly affects the other test in `test_entrypoints.py`, so we might be OK just based on pytest's default ordering of tests, but I don't know that we want to rely on that. Also, there's no guarantee that the monkey-patching wouldn't affect any of numcodecs dependencies in unexpected ways.

If you think those cases aren't worth worrying about, though, then I can revert the multiprocessing change.

jakirkham:
Ah missed we added tests for the backport package `importlib_metadata`

Ok would it be possible to move these into their own test file? Maybe `test_entrypoints_backport.py` or similar? We could then also have some top-level check to make sure `importlib_metadata` is around before testing (ideally without `import`ing given the associated issues)

vyasr:
Done. The test is now in its own module and only runs if `importlib_metadata` is available.

numcodecs/tests/test_entrypoints_backport.py
```diff
@@ -0,0 +1,32 @@
...
+
+def get_entrypoints_with_importlib_metadata_loaded():
+    # importlib_metadata patches importlib.metadata, which can lead to breaking changes
+    # to the APIs of EntryPoint objects used when registering entrypoints. Attempt to
+    # isolate those changes to just this test.
+    import importlib_metadata  # noqa: F401
+    sys.path.append(here)
+    numcodecs.registry.run_entrypoints()
+    cls = numcodecs.registry.get_codec({"id": "test"})
+    assert cls.codec_id == "test"
```
jakirkham, line 25: Would it make sense to reuse `set_path` here somehow? Should cleanup `sys.path` and the `numcodecs.registry` after?

numcodecs/tests/test_entrypoints_backport.py
```diff
@@ -0,0 +1,32 @@
...
+import pkgutil
+import sys
+
+import pytest
+
+from multiprocessing import Process
+
+import numcodecs.registry
+
+if not pkgutil.find_loader("importlib_metadata"):
```
jakirkham, line 11:
```suggestion
if not pkgutil.find_loader("importlib_metadata"):  # pragma: no cover
```

docs/release.rst
```diff
@@ -16,6 +16,8 @@ Fix
 
 * ``Codec`` is now derived from ``abc.ABC``
   By :user:`Mads R. B. Kristensen <madsbk>`, :issue:`472`.
+* Fix handling of entry points on older Python versions where importlib_metadata compatibility is concerned
```
jakirkham, line 19:
```suggestion
* Fix handling of entry points on older Python versions where ``importlib_metadata`` compatibility is concerned
```

numcodecs/tests/test_entrypoints_backport.py
vyasr: I thought about reusing, but I think it'd be more work than it's worth. We can't use a fixture directly because it's not supposed to affect the actual test function, we need it to affect the helper function that's run in the subprocess and that function isn't a test so pytest won't patch in fixtures. We could try and split it into a helper function like I had initially, but it's not really necessary. Since the function is running in a subprocess we don't need to clean things up, it shouldn't affect the modules in the parent process.

jakirkham:
Thanks Vyas! 🙏

PR #477, opened 2023-10-10
Make shuffle pyx functions `noexcept`

### LINKED ISSUE #476: `_doShuffle`: Cython performance hint regarding GIL usage
Am seeing the following performance hint from Cython regarding GIL usage in `_doShuffle` (due to exception raising). Would be good to address

```cython
performance hint: numcodecs/_shuffle.pyx:11:6: Exception check on '_doShuffle' will always require the GIL to be acquired. Possible solutions:
	1. Declare the function as 'noexcept' if you control the definition and you're sure you don't want the function to raise exceptions.
	2. Use an 'int' return type on the function to allow an error code to be returned.
performance hint: numcodecs/_shuffle.pyx:23:6: Exception check on '_doUnshuffle' will always require the GIL to be acquired. Possible solutions:
	1. Declare the function as 'noexcept' if you control the definition and you're sure you don't want the function to raise exceptions.
	2. Use an 'int' return type on the function to allow an error code to be returned.
```

martindurant:
Fixes #476 

@jakirkham , like this? The functions don't actually return anything. 

Do we have any process to determine which codecs release the GIL? Given the performance discussions, this could be imporant.

### DIFF
--- a/numcodecs/_shuffle.pyx
+++ b/numcodecs/_shuffle.pyx
@@ -8,7 +8,7 @@ cimport cython
 
 @cython.boundscheck(False)
 @cython.wraparound(False)
-cpdef void _doShuffle(const unsigned char[::1] src, unsigned char[::1] des, Py_ssize_t element_size) nogil:
+cpdef int _doShuffle(const unsigned char[::1] src, unsigned char[::1] des, Py_ssize_t element_size) nogil:
     cdef Py_ssize_t count, i, j, offset, byte_index
     count = len(src) // element_size
     for i in range(count):
@@ -20,7 +20,7 @@ cpdef void _doShuffle(const unsigned char[::1] src, unsigned char[::1] des, Py_s
 
 @cython.boundscheck(False)
 @cython.wraparound(False)
-cpdef void _doUnshuffle(const unsigned char[::1] src, unsigned char[::1] des, Py_ssize_t element_size) nogil:
+cpdef int _doUnshuffle(const unsigned char[::1] src, unsigned char[::1] des, Py_ssize_t element_size) nogil:
     cdef Py_ssize_t count, i, j, offset, byte_index
     count = len(src) // element_size
     for i in range(element_size):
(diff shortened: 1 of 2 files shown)


numcodecs/_shuffle.pyx
```diff
@@ -8,7 +8,7 @@ cimport cython
 
 @cython.boundscheck(False)
 @cython.wraparound(False)
-cpdef void _doShuffle(const unsigned char[::1] src, unsigned char[::1] des, Py_ssize_t element_size) nogil:
+cpdef int _doShuffle(const unsigned char[::1] src, unsigned char[::1] des, Py_ssize_t element_size) nogil:
```
jakirkham, line 11:
```suggestion
cpdef void _doShuffle(const unsigned char[::1] src, unsigned char[::1] des, Py_ssize_t element_size) noexcept nogil:
```

numcodecs/_shuffle.pyx
```diff
@@ -20,7 +20,7 @@ cpdef void _doShuffle(const unsigned char[::1] src, unsigned char[::1] des, Py_s
 
 @cython.boundscheck(False)
 @cython.wraparound(False)
-cpdef void _doUnshuffle(const unsigned char[::1] src, unsigned char[::1] des, Py_ssize_t element_size) nogil:
+cpdef int _doUnshuffle(const unsigned char[::1] src, unsigned char[::1] des, Py_ssize_t element_size) nogil:
```
jakirkham, line 23:
```suggestion
cpdef void _doUnshuffle(const unsigned char[::1] src, unsigned char[::1] des, Py_ssize_t element_size) noexcept nogil:
```

jakirkham (commented):
Thanks Martin! 🙏

IIUC adding `noexcept` should be sufficient here (looking at [this Cython doc section]( http://docs.cython.org/en/latest/src/userguide/nogil.html#marking-functions-as-able-to-run-without-the-gil ))

jakirkham:
Thanks Martin! 🙏

jakirkham (approved):

PR #489, opened 2023-11-25
DEV: add Cython generated files to gitignore

tacaswell:
To make sure they do not accidentally get re-committed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -101,3 +101,6 @@ ENV/
 
 # setuptools-scm
 numcodecs/version.py
+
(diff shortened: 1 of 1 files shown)


martindurant:
Are they currently tracked? If yes, we should remove them, else I think the old versions will persist.

joshmoore:
Thanks, @tacaswell!

joshmoore (approved):

jakirkham:
Thanks all! 🙏

PR #501, opened 2024-01-24
Add pcodec

rabernat:
This PR adds an exciting new codec to numcodecs: [pcodec](https://github.com/mwlon/pcodec).

From the point of view of the V3 [codec spec](https://zarr-specs.readthedocs.io/en/latest/v3/core/v3.0.html#id18) pcodec is an `array -> bytes`. It takes a numpy array of dtype i4, i8, u4, u8, f4, f8 and turns it into encoded bytes (and reverses this for decoding). It seems like numcodecs hasn't quite formalized this way of categorizing codecs (e.g. there is no `numcodecs.abc.ArrayToBytesCodec`). I've done my best to apply appropriate testing.

TODO:

- [x] Tests pass locally
- [x] Docs build locally


numcodecs/pcodec.py
```diff
@@ -0,0 +1,77 @@
...
+        # note that we use `level` instead of `compression_level` to
+        # match other codecs
+        self.level = level
+        self.delta_encoding_order = delta_encoding_order
+        self.int_mult_spec = int_mult_spec
+        self.float_mult_spec = float_mult_spec
+        self.max_page_n = max_page_n
+
+    def encode(self, buf):
+        print("first", buf.dtype)
```
martindurant, line 59:
```suggestion
```

numcodecs/pcodec.py
```diff
@@ -0,0 +1,77 @@
...
+        buf = ensure_contiguous_ndarray(buf)
+        print("second", buf.dtype)
```
martindurant, line 61:
```suggestion
```

martindurant (commented):
That's a lot of test files!

numcodecs/pcodec.py
rabernat: Oops, thanks nice catch.

rabernat:
> That's a lot of test files!

The test files were auto generated. Do you think this is a problem?

martindurant:
If you can generate them reliably at CI/test time even better; but they are not big, so I am not worried.

rabernat:
> If you can generate them reliably at CI/test time even better

My understanding is that the reason we store files in the repo is to ensure backwards compatibility of the codecs, i.e. to guard against changes to the implementation that would render existing data unreadable. That wouldn't work if you regenerate them each time.

martindurant:
> That wouldn't work if you regenerate them each time.

Right, it would need to be a totally independent and repeatable way to make them. If there is no such way, you store them.

rabernat:
Would anyone care to review this PR?

numcodecs/__init__.py
```diff
@@ -115,3 +115,7 @@
 
 from numcodecs.fletcher32 import Fletcher32
 register_codec(Fletcher32)
+
+with suppress(ImportError):
```
martindurant, line 119: Isn't it better to make sure this import succeeds (`import pcodec` inside `encode`/`decode`, below), but presenting a helpful message whenever someone tries to use Pcodec without the necessary deps? Otherwise, the user would just see some "codec not found" message without any idea of what to do about it.

numcodecs/pcodec.py
```diff
@@ -0,0 +1,78 @@
...
+    ):
+        # note that we use `level` instead of `compression_level` to
+        # match other codecs
+        self.level = level
+        self.delta_encoding_order = delta_encoding_order
+        self.int_mult_spec = int_mult_spec
+        self.float_mult_spec = float_mult_spec
+        self.max_page_n = max_page_n
+
+    def encode(self, buf):
```
martindurant, line 61:
```suggestion
    def encode(self, buf) -> bytes:
```
?
(This may be a good way to label array-bytes codecs; maybe also type for `buf` should be `ndarray`)

numcodecs/__init__.py
rabernat: I agree with this comment. The approach used here was copied from other implementations, but I'm going to change it to what you suggested.

numcodecs/pcodec.py
rabernat: I'm not convinced that there is much value in adding these sorts of type hints if we are not actually running type checking on the library.

martindurant: Is it not for other users of the library, and their IDEs?

rabernat: But incorrect type hints are worse than none at all! For example, is `ndarray` really the correct type for buf? Maybe, but who knows? I could add it, but without running mypy, we'll never know for sure.

martindurant: True that correctness is important, of course; but this is like the "light" version of array->array V array->bytes V bytes->bytes. Still useful. You can always get around mypy too, if you want.

rabernat:
I am sad that the CI is now failing. I don't understand why. (Everything is good locally).

Going to try reverting the change to how we import.

PR #519, opened 2024-04-22
Adds checksum flag to zstd codec

normanrz:
This PR adds the `checksum` flag to the `zstd` codec. This is necessary to support [the proposed `zstd` codec for Zarr3](https://github.com/zarr-developers/zarr-specs/pull/256/files). We need it for the v3 refactoring of zarr-python.

TODO:

- [x] Tests pass locally
- [x] Docs build locally


normanrz:
I would love a review on this so that we can ship this in the upcoming zarr-python 3 release.

d-v-b:
@mkitti would you mind giving this a look?

martindurant:
A good time to switch to cramjam? Does this zstd provide anything that that one doesn't?

normanrz:
> A good time to switch to cramjam? Does this zstd provide anything that that one doesn't?

`cramjam` also does not expose the `checksum` option.

mkitti:
I'm looking. We should normalize the implementation here such that negative compression levels are passed on to the C library and that the default compression level is the default compression level of the underlying C library.

mkitti:
The default CLEVEL here should be changed to `0`. That should be passed on the C library directly without further logic.

`DEFAULT_CLEVEL = 0`

normanrz:
> The default CLEVEL here should be changed to `0`. That should be passed on the C library directly without further logic.
> 
> `DEFAULT_CLEVEL = 0`

I made that change. This is a breaking change, though.

mkitti:
> I made that change. This is a breaking change, though.

The change is that we used to default to compression level 1 explicitly, overriding the configuration of the underlying Zstandard C library.

Now we pass `0` as the compression level, which uses the default compression level of the C library. The default compression level is usually `3`.

For reference sew
https://facebook.github.io/zstd/zstd_manual.html#Chapter5

normanrz:
> Do we provide access to `ZSTD_defaultCLevel(void)`?

We do not. For what would that be useful?

normanrz:
Are there any objections for changing the default compression level from 1 to 3 (i.e. default of the underlying zstd c library) @zarr-developers/python-core-devs?

normanrz:
But that would mean that people have a modified version of numcodecs with a modified zstd library within blosc. Given that a main purpose of numcodecs is to deliver wheels, that seems pretty unlikely to me.

mkitti:
Note that conda-forge builds Blosc with an independently configured zstandard library:
https://github.com/conda-forge/blosc-feedstock/blob/main/recipe%2Fmeta.yaml#L27

mkitti:
Currently, we depend on Zstd frame content size being  encoded. However, we do know what the uncompressed size of a chunk should be from array information. Perhaps we should not generate an error if the frame content size is not "known" to Zstandard because it is, in fact, known to us.

The frame content size may not be encoded if the encoder is using streaming mode and did not pledge the full size of the content in beginning.

normanrz:
> Currently, we depend on Zstd frame content size being encoded. However, we do know what the uncompressed size of a chunk should be from array information. Perhaps we should not generate an error if the frame content size is not "known" to Zstandard because it is, in fact, known to us.
> 
> The frame content size may not be encoded if the encoder is using streaming mode and did not pledge the full size of the content in beginning.

I think that change would be out of scope for this PR. We should continue this discussion in a new issue.

Issue #502, opened 2024-01-24
Supporting Zarr-Python 3 Codec API


jhamman:
Over in Zarr-Python, we are working on a a new major version (v3). This version will have a slightly [new Codec API](https://github.com/zarr-developers/zarr-python/blob/v3/src/zarr/v3/abc/codec.py) and will expose a new set of codec types (`ArrayArrayCodec`, `ArrayBytesCodec`, `BytesBytesCodec`, etc.). These codec classes are not wildly different from the existing Numcodecs API (perhaps except for the partial encode/decode options) but they are not in perfect alignment. With this in mind, some questions for discussion:

1. Can Numcodecs conform to the Zarr-Python API?
    - I don't think this has to be seen as a breaking change but if it is, how do we weight the potential costs and benefits? 
3. Would Numcodecs be able to register codecs via the Zarr-Python entrypoint mechanism? (e.g. https://github.com/zarr-developers/zarr-python/pull/1588)

jhamman:
cc @zarr-developers/python-core-devs

normanrz:
I fixed the OP.

jhamman:
We discussed this in the Zarr-Python refactor meeting today. The outstanding task here is to experiment with Zarr v3 codec API by exposing this library's compression codecs and pre-compression filters through the `BytesBytesCodec` and `ArrayArrayCodec` interfaces. If that can be done effectively, these can be registered through the entrypoint mechanism described above.

This would be a good project for someone interested in getting involved in zarr-python 3's development.

martindurant:
I would add here, that I think having some fallback support for numcodecs (+ other packages that make codecs following its API) is important to maintain readability of datasets in v3. We need to decide whether we can assume they work on bytes - which is by far the most common case - or otherwise can tell from the signature (or try/except) if they accept/produce arrays. That doesn't seem to hard.

Question: codecs are of course CPU-bound, and will be run in threads, hoping that the GIL is released. The to_thread call lives in zarr-python?

If all this is true, I don't see any reason to rewrite any codecs for v3, except where we wish to state the bytes Vs array nature of a codec.

normanrz:
There is now a PR that adds the `numcodecs.zarr3` module which contains Zarr v3 wrappers for the numcodecs codecs: #524

dstansby:
Removing the "good first issue" label since there's now a PR for this

PR #524, opened 2024-05-06
Add wrappers for zarr v3

normanrz:
The Zarr v3 specification only lists a few codecs that are officially supported. However, it is desirable to expose the codecs in numcodecs for use with v3 arrays as well. This PR adds wrapper classes for numcodecs support.

The name of the codecs is prefixed with `numcodecs.` to avoid naming collisions in case some codecs of numcodecs get added to the Zarr spec. Also, there is a warning that numcodecs codecs are not officially supported and will likely not work in any other Zarr implementation.

Most array-to-array ("filters") and bytes-to-bytes codecs are supported. Absent are the variable-length codecs as well as json, msgpack and pickle.

Here is an example of the persisted configuration:
```bash
{
  "name": "numcodecs.fixedoffsetscale",
  "configuration": {"offset": 0, "scale": 51, "astype": "uint16"}
}
```

Use of numcodecs in v2 arrays is not affected.

Fixes https://github.com/zarr-developers/numcodecs/issues/502


rabernat:
> The name of the codecs is prefixed with `https://zarr.dev/numcodecs/` to avoid naming collisions in case some codecs of numcodecs get added to the Zarr spec

I am not sure about the idea of using a URL that does not actually resolve to anything useful.

rabernat:
pcodec is actually an "Array to Bytes" codec: https://github.com/zarr-developers/numcodecs/blob/main/numcodecs/pcodec.py

How would that fit in here?

martindurant:
Any thoughts about what to do with numcodecs codecs not defined in this repo, but currently used via entrypoints?

d-v-b:
> > The name of the codecs is prefixed with `https://zarr.dev/numcodecs/` to avoid naming collisions in case some codecs of numcodecs get added to the Zarr spec
> 
> I am not sure about the idea of using a URL that does not actually resolve to anything useful.

seconding this sentiment, a URL that doesn't resolve to anything is rather confusing. I think `numcodecs.<codec_name>` or `numcodecs/<codec_name>`  are simpler templates for a numcodecs-qualified name.

rabernat:
> Any thoughts about what to do with numcodecs codecs not defined in this repo, but currently used via entrypoints?

Could we ask those codecs to implement Zarr codec entrypoints directly? Which codecs do you have in mind?

The challenge is that the V3 codecs are quite a bit more explicit in their typing (Array to Bytes, Bytes to Bytes, etc.) than legacy numcodecs codecs. So automatically translating an arbitrary numcodecs codec to a V3 codec is not possible.

martindurant:
I am thinking of https://github.com/fsspec/kerchunk/blob/main/kerchunk/codecs.py and imagecodecs. There are probably others.

normanrz:
> > The name of the codecs is prefixed with `https://zarr.dev/numcodecs/` to avoid naming collisions in case some codecs of numcodecs get added to the Zarr spec
> 
> 
> 
> I am not sure about the idea of using a URL that does not actually resolve to anything useful. 

I had asked @MSanKeys963 to setup the respective redirects to the numcodecs docs. That should solve that.

normanrz:
> pcodec is actually an "Array to Bytes" codec: https://github.com/zarr-developers/numcodecs/blob/main/numcodecs/pcodec.py
> 
> 
> 
> How would that fit in here?

Must have missed pcodec. I'll add it.

dstansby:
Could you say something a bit about why this code makes sense to be in `numcodecs` instead of `zarr-python`? My first thoughts are that `numcodecs` sees much less development/maintenance than `zarr-python` at the moment, so unless there's a good reason maybe this code should live in `zarr-python`?

normanrz:
The idea was that the `zarr` package contains the "official" Zarr3 codecs and other libraries can add other codecs via the entrypoint mechanism. The `zarr` package provides base classes for doing so.
`numcodecs` is one of these libraries that can provide additional codecs. There is quite a bit of glue code to make the existing v2 codecs work with the new v3 base classes. This glue code is tightly coupled with the codecs itself, which is why I think it makes more sense to have it in `numcodecs` rather than `zarr`. If we would add a new codec to `numcodecs`, we would need to make a new `zarr` release to support it.
I see that it is a bit weird because `zarr` itself depends on `numcodecs`.

dstansby:
That makes sense - I'll try and give this a proper review in the next couple of days!

numcodecs/tests/test_zarr3.py
```diff
@@ -0,0 +1,206 @@
...
+    from zarr.abc.store import Store
+    from zarr.store import MemoryStore, StorePath
+
+except ImportError:
+    pass
+
+
+pytestmark = pytest.mark.skipif(
+    sys.version_info < (3, 10), reason="zarr-python 3 requires Python 3.10 or higher"
+)
```
dstansby, line 25: Don't need this, since `numcodecs` requires 3.10 or higher too

numcodecs/tests/test_zarr3.py
```diff
@@ -0,0 +1,206 @@
...
+try:
+    from zarr.codecs.registry import get_codec_class
+    from zarr.array import Array
+    from zarr.common import JSON
+    from zarr.codecs import BytesCodec
+    from zarr.abc.store import Store
+    from zarr.store import MemoryStore, StorePath
+
+except ImportError:
+    pass
```
dstansby, line 20: Should this whole try; except block be replaced by a pytest skip that skips if zarr<3 is installed?

numcodecs/tests/test_zarr3.py
```diff
@@ -0,0 +1,206 @@
...
+)
+
+
+@pytest.fixture
+def store() -> Iterator[Store]:
+    yield StorePath(MemoryStore(mode="w"))
+
+
+@pytest.mark.parametrize(
+    "codec_id", ["blosc", "lz4", "zstd", "zlib", "gzip", "bz2", "lzma", "shuffle"]
```
dstansby, line 34: These codecs are specifically the ones that zarr v3 needs support for? If so it might be more obvious if there's a `ZARR3_CODECS` list defined near the top of this file with this list.

numcodecs/tests/test_zarr3.py
```diff
@@ -0,0 +1,206 @@
...
+)
+def test_generic_codec(store: Store, codec_id: str):
+    data = np.arange(0, 256, dtype="uint16").reshape((16, 16))
+
+    with pytest.warns(UserWarning, match="Numcodecs.*"):
```
dstansby, line 39: Can you make this match more specific? It's not clear what is's warning on just from Numcodecs.*

numcodecs/tests/test_zarr3.py
```diff
@@ -0,0 +1,206 @@
...
+            dtype=data.dtype,
+            fill_value=0,
+            codecs=[
+                BytesCodec(),
+                get_codec_class(f"numcodecs.{codec_id}")({"id": codec_id}),
+            ],
+        )
+
+    a[:, :] = data.copy()
+    assert np.array_equal(data, a[:, :])
```
dstansby, line 53:
```suggestion
   np.testing.assert_array_equal(data, a[:, :])
```
I think will give more information if the test fails?

numcodecs/zarr3.py
```diff
@@ -0,0 +1,275 @@
...
+
+from dataclasses import dataclass, replace
+from functools import cached_property
+import math
+from typing_extensions import Self
+from warnings import warn
+
+import numpy as np
+import numcodecs
+
```
dstansby, line 11: Might be nice to have a try: import zarr; assert zarr.version=3; except: raise ImportError("zarr is required to use numcodecs zarr integration block (sorry for my awful pseudocode), so there's a nice error if a user tries to import but doesn't have zarr>=3 installed.

numcodecs/zarr3.py
```diff
@@ -0,0 +1,275 @@
...
+        )
+    id = parsed_name[slice(len(expected_name_prefix), None)]
+    return {"id": id, **parsed_configuration}
+
+
+@dataclass(frozen=True)
+class NumcodecsCodec:
+    codec_config: dict[str, JSON]
+
+    def __init__(self, *, codec_id: str | None = None, codec_config: dict[str, JSON]) -> None:
```
dstansby, line 41: Can the codec_id type be narrowed to a Literal list of supported codecs?

dstansby (commented):
I haven't quite looked at everything, but here's some initial comments

numcodecs/tests/test_zarr3.py
normanrz: Not really. These are all the codecs that have a generic-enough config that we can use one parameterized test. Other codecs need specialized tests.

numcodecs/tests/test_zarr3.py
```diff
@@ -0,0 +1,203 @@
...
+from collections.abc import Iterator
+
+
+import numpy as np
+import pytest
+
+from numcodecs.registry import get_codec
+
+
+zarr = pytest.importorskip("docutils")
```
dstansby, line 11:
```suggestion
zarr = pytest.importorskip("zarr")
```
?

.github/workflows/ci.yaml
```diff
@@ -63,6 +63,12 @@ jobs:
           # TODO: put back zfpy import when it supports numpy 2.0
           python -m pip install -v -e .[test,test_extras,msgpack,pcodec]
 
+      - name: Install zarr-python
+        shell: "bash -l {0}"
+        run: |
+          conda activate env
+          python -m pip install zarr==3.0.0a0
```
dstansby, line 70: We need a way to run tests on both zarr v2 and zarr v3 before this is merged.

numcodecs/zarr3.py
```diff
@@ -0,0 +1,282 @@
...
+from typing_extensions import Self
+from warnings import warn
+
+import numpy as np
+import numcodecs
+
+try:
+    import zarr
+
+    assert zarr.__version__ >= "3.0.0"
```
dstansby, line 15: There should be a test that this errors as expected (a nice import error) if zarr v2 is installed.

dstansby (changes requested):
I haven't reviewed the code in detail because I haven't got the time to get the context on zarr v3 I need to do that. I've left some comments on what I'd like to see addressed before this is merged though, mainly to ensure we maintain compatibility with zarr v2.

I also wonder if we should make all of the `zarr3` api private, just exposing it via the entrypoints, to give more flexibility for modifying it in the future?

.github/workflows/ci.yaml
normanrz: Just had another look. The `zarr` package is only used for these Zarr3 codecs and their corresponding tests. Previously, `zarr` wasn't even installed in the CI. I don't think we need to test against zarr-python 2, here.

normanrz:
@dstansby I incorporated your feedback:
- `zarr3.py` -> `_zarr3.py`
- Added a test for trying to import zarr-python 2. I had to add another CI matrix run for installing zarr-python 2. Not sure if this test is worth the extra CI time.

Not sure what is going on with codecov. I think all lines should be tested by at least a few CI runs. Help appreciated.

.github/workflows/ci.yaml
dstansby: Ahhh that makes sense, don't know why I got confused!

dstansby:
Making it public sounds good - the API reference will need adding to https://numcodecs--524.org.readthedocs.build/en/524/api.html if it's going to be public. I'll have a look and try and debug the codecov issue now.

.github/workflows/ci.yaml
```diff
@@ -15,6 +15,7 @@ jobs:
         python-version: ["3.10", "3.11", "3.12", "3.13.0"]
         # macos-12 is an intel runner, macos-14 is a arm64 runner
         platform: [ubuntu-latest, windows-latest, macos-12, macos-14]
+        zarr-version: ["zarr>=2,<3", "zarr==3.0.0b0"]
```
dstansby, line 18: Sorry if I lost track of this - I think you convinced me earlier in the comments of this PR that we could just install zarr==3 here and not bother testing with zarr v2?

dstansby:
The numcodecs issue is because the version check isn't working. If you check the logs of the runs with zarr v3, they contain the line

```
SKIPPED [2] numcodecs/tests/test_zarr3.py:11: module 'zarr' has __version__ '3.0.0b0', required is: '3.0.0'
```

.github/workflows/ci.yaml
normanrz: It is only there for this test: https://github.com/zarr-developers/numcodecs/pull/524/files#diff-b3efef438b6c87206e5b2229310709c6b066d5c41d3c8748c66b8c9ddb0d5ed8R13-R14
Might not be worth it.

dstansby:
Just to check, I think the final thing to be done here is adding some documentation? (API, which probably includes adding docstrings and a short description of the purpose of the new sub-module)

normanrz:
Yes. I'll write some API docs. And also add the new crc32c codec.

normanrz:
I now added docs for this module. 
Screenshot 2024-11-05 at 18 11 10

I also stripped the `Codec` suffix of all the wrapper codecs so that they are called the same as the original codecs. I think this is a nice way of interacting with the codecs. As part of that, I also reworked the tests.

I deactivated zfpy in macos-14 builds. zfpy for arm64 seems to specify other numpy versions which clashes with the numpy requirements of zarr3.

dstansby (commented):
Sorry, I won't have time for a full review for a few days, but hoping this partial one is helpful.

The build docs aren't rendering properly: https://numcodecs--524.org.readthedocs.build/en/524/zarr3.html

Looking at the [readthedocs build log](https://app.readthedocs.org/projects/numcodecs/builds/26183071/) this is because
1. zarr-python v3 isn't being installed, which is required to import `numcodecs.zarr` and read the docstring.
2. The codec names clash, so need fully specifying as e.g. `numcodecss.zarr3.CodecName` instead of just `CodecName`

I'm happy to put together a fix for the doc issues when I have time, but might not be for a couple of days.

.github/workflows/ci.yaml
```diff
@@ -70,10 +70,15 @@ jobs:
             conda activate env
             python -m pip install -v ".[pcodec]"
 
+      - name: Install zarr-python
+        shell: "bash -l {0}"
+        run: |
+          conda activate env
+          python -m pip install 'zarr==3.0.0b1'
```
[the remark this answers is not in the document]
normanrz: Actually, I had to use `python -m pip install --pre zarr`

normanrz:
Thanks for checking the docs CI. I didn't catch that. I fixed that right now: https://numcodecs--524.org.readthedocs.build/en/524/zarr3.html

The zarr3 codecs link back to the original codecs. Should I also add links the other way round (e.g. numcodecs.blosc.Blosc -> numcodecs.zarr3.Blosc)?

.github/workflows/ci.yaml
```diff
@@ -70,10 +70,15 @@ jobs:
...
 
       # This is used to test with zfpy, which does not yet support numpy 2.0
       - name: Install older numpy and zfpy
-        if: matrix.python-version == '3.11'
+        if: matrix.python-version == '3.11' && matrix.platform != 'macos-14'
```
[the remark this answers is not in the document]
normanrz: I just approved that PR. Let's get that merged first.

numcodecs/zarr3.py
```diff
@@ -0,0 +1,379 @@
...
+
+CODEC_PREFIX = "numcodecs."
+
+
+def _expect_name_prefix(codec_name: str) -> str:
+    if not codec_name.startswith(CODEC_PREFIX):
+        raise ValueError(
+            f"Expected name to start with '{CODEC_PREFIX}'. Got {codec_name} instead."
+        )  # pragma: no cover
+    return codec_name[len(CODEC_PREFIX) :]
```
dstansby, line 61:
```suggestion
    return codec_name.removeprefix(CODEC_PREFIX)
```
A bit clearer?

numcodecs/zarr3.py
```diff
@@ -0,0 +1,379 @@
...
+
+    def to_dict(self) -> JSON:
+        codec_config = self.codec_config.copy()
+        return {
+            "name": self.codec_name,
+            "configuration": codec_config,
+        }
+
+    def compute_encoded_size(self, input_byte_length: int, chunk_spec: ArraySpec) -> int:
+        return input_byte_length  # pragma: no cover
```
dstansby, line 117:
```suggestion
        raise NotImplementedError  # pragma: no cover
```
Would it be better to raise here since this class is intended to be overidden?

dstansby (approved):
LGTM 👍 - I've left two optional comments, feel free to take them or leave them and then self merge

dstansby:
> The zarr3 codecs link back to the original codecs. Should I also add links the other way round (e.g. numcodecs.blosc.Blosc -> numcodecs.zarr3.Blosc)?

I don't think it's worth doing this

normanrz:
This PR got merged prematurely. Looks like we should better set up the auto-merge.

Anyways, this test is still failing after merging #620. Should I add https://github.com/zarr-developers/numcodecs/pull/524/files/71178b07b04099f9a5953f2513599a95ae810c0f#diff-944291df2c9c06359d37cc8833d182d705c9e8c3108e7cfe132d61a06e9133dd back @dstansby?

Issue #521, opened 2024-04-25
NumPy 2.0 support


jakirkham:
NumPy 2.0 is coming out soon ( https://github.com/numpy/numpy/issues/24300 ). NumPy 2.0.0rc1 packages for conda & wheels came out 3 weeks back ( https://github.com/numpy/numpy/issues/24300#issuecomment-2030603395 )

To prepare for NumPy 2.0, it might be worthwhile to start testing against NumPy 2 in CI

NumPy has put out [a migration guide]( https://numpy.org/devdocs/numpy_2_0_migration_guide.html#numpy-2-migration-guide ). More details are in [the release notes]( https://numpy.org/devdocs/release/2.0.0-notes.html )

If Numcodecs make use of NumPy's C API (and produces wheels that use it), having a release of Numcodecs with wheels built against NumPy 2.0.0rc1 would be helpful to ensure NumPy 1 & 2 compatible wheels (as wheels built against NumPy 1 won't be compatible with NumPy 2). More details in [this NumPy 2 ABI doc]( https://numpy.org/devdocs/dev/depending_on_numpy.html#numpy-2-abi-handling )

Also as NumPy is tracking ecosystem support for NumPy 2.0, it would be helpful to share Numcodecs current support status in issue (with any plans): https://github.com/numpy/numpy/issues/26191

jakirkham:
As Numcodecs has relied primarily on the Python Buffer Protocol and not really used NumPy's C API, this shouldn't involve rebuilding Numcodecs. Have just searched through the code to reconfirm this

The thing to check for would be if Numcodecs is affected by any NumPy Python API breaks. This could be done by adding it to CI here

jakirkham:
FWIW attempted running the test suite locally with NumPy 2.0.0rc1 from conda-forge and it appeared to pass. So doesn't look like any changes are needed

Still would be good to add to CI

jakirkham:
We do have an optional dependency on zfpy

https://github.com/zarr-developers/numcodecs/blob/4abe4be2d32a1e5743982f3056aac937f9664da6/pyproject.toml#L66-L68

zfpy still needs to do some work to migrate: https://github.com/LLNL/zfp/issues/210

However this just means users won't be able to install zfpy until that is resolved. No impact on installing Numcodecs itself

dstansby:
It might be worth doing a release with the `numpy<2` pin first (which is now in `main`) while this gets sorted out?

joshmoore:
[...]

PR #531, opened 2024-05-24
Bump min Python to 3.10

dstansby:
As per [SPEC 0](https://scientific-python.org/specs/spec-0000/) (which it says that zarr is signed up to), drop support from Python 3.8/3.9.

### DIFF
--- a/.github/workflows/ci-linux.yaml
+++ b/.github/workflows/ci-linux.yaml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.10", "3.11", "3.12"]
 
     steps:
       - name: Checkout source
--- a/.github/workflows/ci-osx.yaml
+++ b/.github/workflows/ci-osx.yaml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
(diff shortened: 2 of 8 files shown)


martindurant:
This would be a good time to ensure numpy 2 support, if it isn't already done.

dstansby:
There's an issue here: https://github.com/zarr-developers/numcodecs/issues/521; I'm happy to take a look, but don't want to open too many PRs at once 😄

normanrz:
It would be great if you could have a look at numpy 2 support.

I wonder if we should remove line of code that codecov complains about? https://app.codecov.io/gh/zarr-developers/numcodecs/pull/531/indirect-changes?src=pr&el=tree-more&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=zarr-developers

dstansby:
This is good to go - the yellow CI jobs are ones no longer run as they're old versions of Python. I think removing these requires some manual adjustment of repository settings from an admin.

normanrz:
I removed the requirement for py 3.8 and 3.9. Codecov still needs a bump.

normanrz (approved):

dstansby:
From one of the builds:

> [2024-06-19T09:51:18.726Z] ['error'] There was an error running the uploader: Error uploading to [https://codecov.io:](https://codecov.io/) Error: There was an error fetching the storage URL during POST: 429 - {'detail': ErrorDetail(string='Rate limit reached. Please upload with the Codecov repository upload token to resolve issue. Expected time to availability: 2676s.', code='throttled')}

So I think someone needs to set up using a codecov token for codecov to start working again.

PR #535, opened 2024-06-19
Add numpy 2.0 compatibility

dstansby:
Fixes https://github.com/zarr-developers/numcodecs/issues/521.

Currently the solution is to just test without zfpy support, which is easy to implement. In an ideaL world I guess we want to run at least one test with zfpy and `numpy<2`? I won't have time to implement that in the next few days, so someone else feel free to take over and do that if desirable.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -62,6 +62,14 @@ jobs:
           # TODO: put back zfpy import when it supports numpy 2.0
           python -m pip install -v -e .[test,test_extras,msgpack,pcodec]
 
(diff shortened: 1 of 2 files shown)


normanrz (approved):
I think this it a good solution. I will merge tomorrow if there are no objections.

We should track [numpy 2.0 support in zfpy](https://github.com/LLNL/zfp/pull/231), though.

dstansby:
I updated this so the Python 3.10 tests use `numpy<2` and `zfpy`, so hopefully we still get code coverage for zfpy tests?

PR #536, opened 2024-06-19
Build macOS arm64 wheels on arm64 runner

dstansby:
This should:

- Allow tests to be run for the arm64 wheels
- Speed up the wheel CI job a bit, as the {intel, arm64} macOS builds will now be done in parallel on different jobs.


normanrz:
Should we run the build tests also on macos-14 as well? https://github.com/zarr-developers/numcodecs/actions/runs/9586809165/job/26435499989?pr=536

dstansby:
👍 good idea

normanrz (approved):

PR #540, opened 2024-06-29
Fix ZFPY import error

px39n:
When 

numpy >=2.0.0
from numcodecs.zfpy import ZFPY

raise error [ValueError: numpy.ndarray size changed, may indicate binary incompatibility. Expected 88 from C header, got 80 from PyObject]

The error is caused by zfpy library crashing with recent ABI breaking of numpy udpate. So we'd better force numpy <2.0 unless zfpy update again.

Reference: (https://stackoverflow.com/questions/66060487/valueerror-numpy-ndarray-size-changed-may-indicate-binary-incompatibility-exp)

[Description of PR]

TODO:

- [ ] Tests pass locally
- [ ] Docs build locally

### DIFF
--- a/numcodecs/zfpy.py
+++ b/numcodecs/zfpy.py
@@ -11,7 +11,7 @@ with suppress(PackageNotFoundError):
 if _zfpy_version:
     # Check NumPy version
     _numpy_version: tuple = tuple(map(int, version("numpy").split('.')))
-    if numpy_version_tuple >= (2, 0, 0) and zfpy_version_tuple <= (1, 0, 1):
+    if _numpy_version >= (2, 0, 0) and _zfpy_version <= (1, 0, 1):
         _zfpy_version = ()
         warnings.warn(
             "NumPy version >= 2.0.0 detected. The zfpy library is incompatible with this version of NumPy. "
--- a/numcodecs/zfpy.py
+++ b/numcodecs/zfpy.py
@@ -16,11 +16,11 @@ if _zfpy_version:
         warnings.warn(
             "NumPy version >= 2.0.0 detected. The zfpy library is incompatible with this version of NumPy. "
             "Please downgrade to NumPy < 2.0.0 or wait for an update from zfpy.",
-            UserWarning
+            UserWarning,
         )
 else:
     with suppress(ImportError):
-        import zfpy as _zfpy 
+        import zfpy as _zfpy
 
 if _zfpy:
     from .abc import Codec
--- a/numcodecs/zfpy.py
+++ b/numcodecs/zfpy.py
@@ -4,16 +4,20 @@ import warnings
 
 _zfpy = None
 
-# Check NumPy version
-numpy_version = np.__version__
-numpy_version_tuple = tuple(map(int, numpy_version.split('.')))
-if numpy_version_tuple >= (2, 0, 0):
-    _zfpy = None
-    warnings.warn(
-        "NumPy version >= 2.0.0 detected. The zfpy library is incompatible with this version of NumPy. "
-        "Please downgrade to NumPy < 2.0.0 or wait for an update from zfpy.",
-        UserWarning
-    )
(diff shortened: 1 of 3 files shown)


normanrz:
We want to support numpy 2.0. Unfortunately, that means we need to wait until zfpy has been updated. See https://github.com/zarr-developers/numcodecs/pull/535

jakirkham:
Agreed

That said, do understand why this error is confusing

Perhaps we can add some logic here to check NumPy version and disable the extension with an appropriate message to the user

https://github.com/zarr-developers/numcodecs/blob/342754d736aa5a0c79024f87e0a5b8d1576c1e28/numcodecs/zfpy.py#L3-L5

We could also ensure that the `zfpy` optional dependency has this constraint

https://github.com/zarr-developers/numcodecs/blob/342754d736aa5a0c79024f87e0a5b8d1576c1e28/pyproject.toml#L65-L67

numcodecs/zfpy.py
```diff
@@ -1,9 +1,22 @@
 from contextlib import suppress
+import warnings
+import numpy as np
```
jakirkham, line 3:
```suggestion
from importlib.metadata import PackageNotFoundError, version
import warnings
```

numcodecs/zfpy.py
```diff
@@ -1,9 +1,22 @@
...
+if numpy_version_tuple >= (2, 0, 0):
+    _zfpy = None
+    warnings.warn(
+        "NumPy version >= 2.0.0 detected. The zfpy library is incompatible with this version of NumPy. "
+        "Please downgrade to NumPy < 2.0.0 or wait for an update from zfpy.",
+        UserWarning
+    )
+else:
+    with suppress(ImportError):
+        import zfpy as _zfpy
```
jakirkham, line 19:
```suggestion
_zfpy_version: tuple = ()
with suppress(PackageNotFoundError):
    _zfpy_version = tuple(map(int, version("zfpy").split(".")))

if _zfpy_version:
    # Check NumPy version
    _numpy_version: tuple = tuple(map(int, version("numpy").split('.')))
    if numpy_version_tuple >= (2, 0, 0) and zfpy_version_tuple <= (1, 0, 1):
        _zfpy_version = ()
        warnings.warn(
            "NumPy version >= 2.0.0 detected. The zfpy library is incompatible with this version of NumPy. "
            "Please downgrade to NumPy < 2.0.0 or wait for an update from zfpy.",
            UserWarning
        )
else:
    with suppress(ImportError):
        import zfpy as _zfpy 
```

jakirkham (commented):
Thanks! 🙏

Am wondering if we should check whether `zfpy` exists before warning

Also wondering if we should limit which `zfpy` versions warn (since at some point this will be fixed)

What do you think? 🙂

jakirkham:
Thanks! 🙏

Please add a release note as well. Perhaps under fix?

https://github.com/zarr-developers/numcodecs/blob/342754d736aa5a0c79024f87e0a5b8d1576c1e28/docs/release.rst?plain=1#L27-L33

px39n:
"pre-commit.ci autofix"

px39n:
pre-commit.ci autofix

normanrz:
This looks like a great solution. Thank you!

px39n:
@jakirkham Thank you for constructive suggestions, and thanks @normanrz for mention #535 

It helps the numpy 2.0 users like me who would like to use zfpy but no clues where is going wrong with import error

normanrz (approved):

rabernat:
It looks like this PR was merged without hitting the coverage target. Now all subsequent PRs to numcodecs are failing that test (see https://github.com/zarr-developers/numcodecs/pull/544#issuecomment-2215898925).

We should fix this.

normanrz:
Fixed in #546

jakirkham:
Thanks all! 🙏

PR #544, opened 2024-07-06
Upgraded pcodec to 0.2 and used new ModeSpec configuration

mwlon:
Upgraded Pcodec to 0.2. The main API change is replacement of individual mode specs with a single spec for all modes.

### DIFF
--- a/numcodecs/pcodec.py
+++ b/numcodecs/pcodec.py
@@ -68,7 +68,7 @@ class PCodec(numcodecs.abc.Codec):
                 mode_spec = ModeSpec.auto()
             case 'classic':
                 mode_spec = ModeSpec.classic()
-            case _:
+            case _:  # pragma: no cover
                 raise ValueError(f"unknown value for mode_spec: {self.mode_spec}")
         paging_spec = PagingSpec.equal_pages_up_to(self.equal_pages_up_to)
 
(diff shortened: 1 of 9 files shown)


mwlon:
It seems the lines the coverage report is complaining about are deliberately unreachable (with correct mypy typing)

numcodecs/pcodec.py
```diff
@@ -60,20 +57,25 @@ def __init__(
...
     def encode(self, buf):
         buf = ensure_contiguous_ndarray(buf)
 
+        match self.mode_spec:
+            case 'auto':
+                mode_spec = ModeSpec.auto()
+            case 'classic':
+                mode_spec = ModeSpec.classic()
+            case _:
+                raise ValueError(f"unknown value for mode_spec: {self.mode_spec}")
```
rabernat, line 72:
```suggestion
            case _:  # pragma: no cover
                raise ValueError(f"unknown value for mode_spec: {self.mode_spec}")
```

rabernat (approved):
Thanks for doing this @mwlon!

My suggestion should fix the coverage.

mwlon:
Now this is confusing

mwlon:
Oh it looks like it's failing due to unrelated changes from the merge (something zfpy). And I can't tell why ruff is complaining with this last commit... it passes locally, and the "Details" page doesn't say what file it's failing on. @rabernat @normanrz are we good to merge this?

rabernat:
Thanks for your patience with our over-eager CI @mwlon! 🙏 

Yes, you're right that #540 broke our coverage test, so we can ignore that. I wish we could fix pre-commit though. I'll see what I can do.

PR #548, opened 2024-07-12
Add release docs for PRs

normanrz:
For #535 and #531

### DIFF
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -11,6 +11,12 @@ Release notes
 Unreleased
 ----------
 
+
+.. _release_0.13.0:
+
+0.13.0
+------
+
 Enhancements
 ~~~~~~~~~~~~
 * Add checksum flag to zstd and sets the default level to 0.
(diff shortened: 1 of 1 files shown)


rabernat:
@normanrz can we also use this PR to bump the new version number from "unreleased" to 0.13.0?

normanrz:
Sure. Is that something we have to do manually or is there some release automation?

rabernat:
Here is the release process: https://numcodecs.readthedocs.io/en/stable/contributing.html#release-procedure

Before releasing, we need to manually update the release notes and also create a new space for "unreleased". So let's just use this PR to do all of that prep. Once it is merged, I will do the release.

docs/release.rst
```diff
@@ -11,9 +11,11 @@ Release notes
 Unreleased
 ----------
 
-Maintenance
```
rabernat, line 14: I think we need new blank sections for Enhancements, Maintenance, etc. under unreleased.

normanrz: Got it.

PR #550, opened 2024-07-14
Only suppress imports of optional external dependencies

dstansby:
This makes the suppression of import errors consistent with which codecs depend on optional external packages. If a codec doesn't depend on any code external to `numcodecs`, then import errors shouldn't be suppressed as they are indicating an issue with `numcodecs` itself.

I looked in `git blame` and there doesn't seem to be any commentary/reason given for these suppressions, and they all seem to have been added in the original PRs where each codec was added.

This came up over at https://github.com/zarr-developers/zarr-python/issues/2032.

### DIFF
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -24,6 +24,14 @@ Fixes
 
 Improvements
 ~~~~~~~~~~~~
+* If an import error is raised when trying to define a codec that is *not*
+  an optional dependency, it is no longer silently caught. Instead it will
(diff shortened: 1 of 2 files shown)


rabernat:
Thanks for this @dstansby! This seems like the right behavior to me. But it's a significant change. I'd want to get @jakirkham's opinion before merging. Perhaps John has some context on the original intended behavior?

PR #555, opened 2024-07-26
Rewrite logic in `vlen` to drop NumPy build dependency

jakirkham:
As noted in issue ( https://github.com/zarr-developers/numcodecs ), `vlen` had problems with read-only data

Internally this occurred because we assigned the input array to `object[:]`, which expects mutability. Ideally we would just replace this with `const object[:]`. However Cython does not support this ( https://github.com/cython/cython/issues/2485 ). Over time various blockers have been resolved on that front, but it is not yet fixed

Fortunately an astute SciPy developer noted that `const <fused_type>[:]` is allowed ( https://github.com/scipy/scipy/pull/18192#pullrequestreview-1359581611 ). We use that same trick here

This in turn should allow us to continue to accept read-only arrays. Though this does so without a NumPy build dependency, which can introduce more complexity to the build process

<hr>

TODO:

- [ ] Tests pass locally
- [ ] Docs build locally

### DIFF
--- a/numcodecs/vlen.pyx
+++ b/numcodecs/vlen.pyx
@@ -7,7 +7,6 @@
 
 import cython
 cimport cython
-from numpy cimport ndarray
 import numpy as np
 from .abc import Codec
 from .compat_ext cimport Buffer
@@ -16,7 +15,7 @@ from .compat import ensure_contiguous_ndarray
 from cpython cimport (PyBytes_GET_SIZE, PyBytes_AS_STRING, PyBytes_Check,
                       PyBytes_FromStringAndSize, PyUnicode_AsUTF8String)
 from cpython.buffer cimport PyBUF_ANY_CONTIGUOUS
-from libc.stdint cimport uint8_t
+from libc.stdint cimport uint8_t, uintptr_t
 from libc.string cimport memcpy
 from ._utils cimport store_le32, load_le32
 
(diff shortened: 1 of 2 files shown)


numcodecs/vlen.pyx
```diff
@@ -28,6 +27,14 @@ cdef extern from "Python.h":
     int PyUnicode_Check(object text)
 
 
+# Hacky workaround for lack of `const object[:]`.
+# Appears hiding `object` in a fused type works.
+# xref: https://github.com/cython/cython/issues/2485
+# xref: https://github.com/scipy/scipy/pull/18192#pullrequestreview-1359581611
+ctypedef fused object_fused_t:
+    object
+    uintptr_t
```
jakirkham, line 36: This is the crux of the workaround

For fusing, used a `uintptr_t`. As `object` is effectively `PyObject*`, fusing with a generic pointer/address type seemed most appropriate

Note this differs from the SciPy approach where they used `double`. Though it shouldn't matter much what the other type is as it is unused

jakirkham: Well as evidenced by CI, this is not enough. Think what we need to do is get the fused types inside some kind of function signature. Then it might work

However looking at the code here. There are multiple places this would need to go

Wonder if the better path would be...

1. Move most of this logic into pure Python (Cython is only needed for a small portion of this code)
2. Refactor the code to reuse common bits across cases
3. Have a small Cython function that deals with the bits that need to be lower-level

dstansby:
As per comments on https://github.com/conda-forge/numcodecs-feedstock/pull/108, I don't think we need this any more

PR #515, opened 2024-03-12
Allow VLenUTF8 to encode a read-only source

ivirshup:
Allow read-only arrays to be encoded with VLenUTF8

* Fixes #514

Arguably this should be tested and potentially fixed for all codecs.

**Question:** Is this implementation fine? I've done very little Cython, so I have no idea if there are any downsides to this.

TODO:

- [x] Unit tests and/or doctests in docstrings
- [x] Tests pass locally
- [ ] Docstrings and API docs for any new/modified user-facing classes and functions
- [ ] Changes documented in docs/release.rst
- [ ] Docs build locally
- [x] GitHub Actions CI passes
- [x] Test coverage to 100% (Codecov passes)

PR #557, opened 2024-08-06
Revert "Allow VLenUTF8 to encode a read-only source"

rabernat:
Reverts zarr-developers/numcodecs#515, as discussed in https://github.com/conda-forge/numcodecs-feedstock/pull/107#issuecomment-2270007816.

### DIFF
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -39,8 +39,7 @@ Enhancements
 
 Fix
 ~~~
-* Fix VLenUTF8 encoding for read-only buffers.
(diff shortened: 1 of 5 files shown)


dstansby:
As per https://github.com/conda-forge/numcodecs-feedstock/pull/107#issuecomment-2283932269 I don't think this is needed, and can be closed?

rabernat:
Thanks @dstansby! If the failed conda build really has nothing to do with this change, then yes, makes sense to close.

Thanks for all your work on resolving this! 🙏

PR #563, opened 2024-08-20
Build aarch64 wheels

dstansby:
This PR:

- Changes the cibuildwheel config to explicitly list the architectures to build for
- Adds `aarch64` to this list for linux builds
- Restricts the number of wheels built on a PR to 1 per major platform (linux, windows, macOS x86_64, macOS arm64)
- Builds all the wheels on release, or when the "build all wheels" label is added to a PR

This speeds up runs on PRs, but allows the flexibiility of doing a full run of wheel building if needed by adding the label.

Fixes https://github.com/zarr-developers/numcodecs/issues/312, fixes https://github.com/zarr-developers/numcodecs/issues/288, and an updated version of https://github.com/zarr-developers/numcodecs/pull/315

TODO:

### DIFF
--- a/.github/workflows/wheel.yaml
+++ b/.github/workflows/wheel.yaml
@@ -8,17 +8,22 @@ concurrency:
 
 jobs:
   build_wheels:
-    name: Build wheel on ${{ matrix.os }}
+    name: Build ${{ matrix.archs }} wheel on ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
         # macos-13 is an intel runner, macos-14 is a arm64 runner
         os: [ubuntu-latest, windows-latest, macos-13, macos-14]
+        archs: ["native"]
+        include:
+          - os: ubuntu-latest
+            archs: "aarch64"
     env:
       CIBW_TEST_COMMAND: python -c "import numcodecs"
       CIBW_BUILD: "cp311-* cp312-* cp313-*"
       CIBW_SKIP: "pp* *-musllinux_* *win32 *_i686 *_s390x"
+      CIBW_ARCHS: ${{ matrix.archs }}
       # note: CIBW_ENVIRONMENT is now set in pyproject.toml
 
     steps:
--- a/.github/workflows/wheel.yaml
+++ b/.github/workflows/wheel.yaml
@@ -8,22 +8,21 @@ concurrency:
 
 jobs:
   build_wheels:
-    name: Build ${{ matrix.archs }} wheel on ${{ matrix.os }}
+    name: Build wheels on ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
         # macos-13 is an intel runner, macos-14 is a arm64 runner
         os: [ubuntu-latest, windows-latest, macos-13, macos-14]
-        archs: ["native"]
-        include:
-          - os: ubuntu-latest
-            archs: "aarch64"
+        build_all:
+          - ${{ (github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags/v')) || contains(github.event.pull_request.labels.*.name, 'test all wheels')}}
     env:
       CIBW_TEST_COMMAND: python -c "import numcodecs"
       CIBW_BUILD: "cp311-* cp312-* cp313-*"
-      CIBW_SKIP: "pp* *-musllinux_* *win32 *_i686 *_s390x"
-      CIBW_ARCHS: ${{ matrix.archs }}
+      CIBW_ARCHS_MACOS: native
+      CIBW_ARCHS_WINDOWS: native
+      CIBW_ARCHS_LINUX: "x86_64 aarch64"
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -24,7 +24,9 @@
 if have_cflags:
     # respect compiler options set by user
     pass
-elif os.name == 'posix':
+elif os.name == 'posix' and os.uname()[4] != 'aarch64':
+    # These flags aren't recognised by gcc on aarch64
+    # (at least when we build the aarch64 wheens on GitHub actions)
```
jakirkham, line 29:
```suggestion
elif os.name == 'posix' and os.uname()[4] != 'aarch64':
    # These instructions are not available on `aarch64`.
    # So there is no need to disable them as they won't show up there.
```

.github/workflows/wheel.yaml
```diff
@@ -18,14 +30,25 @@ jobs:
     env:
       CIBW_TEST_COMMAND: python -c "import numcodecs"
       CIBW_BUILD: "cp310-* cp311-* cp312-* cp313-*"
-      CIBW_SKIP: "pp* *-musllinux_* *win32 *_i686 *_s390x"
+      CIBW_ARCHS_MACOS: native
+      CIBW_ARCHS_WINDOWS: native
+      CIBW_ARCHS_LINUX: "x86_64 aarch64"
```
jakirkham, line 35: So will this build both Linux wheels in the same job?

.github/workflows/wheel.yaml
```diff
@@ -18,14 +30,25 @@ jobs:
     env:
       CIBW_TEST_COMMAND: python -c "import numcodecs"
       CIBW_BUILD: "cp310-* cp311-* cp312-* cp313-*"
-      CIBW_SKIP: "pp* *-musllinux_* *win32 *_i686 *_s390x"
+      CIBW_ARCHS_MACOS: native
+      CIBW_ARCHS_WINDOWS: native
```
jakirkham, line 34: Do we want to try the same thing for Windows ARM?

```suggestion
      CIBW_ARCHS_WINDOWS: "AMD64 ARM64"
```

Basing this off [this doc]( https://cibuildwheel.pypa.io/en/stable/options/#archs )

.github/workflows/wheel.yaml
```diff
@@ -18,14 +30,25 @@ jobs:
     env:
       CIBW_TEST_COMMAND: python -c "import numcodecs"
       CIBW_BUILD: "cp310-* cp311-* cp312-* cp313-*"
-      CIBW_SKIP: "pp* *-musllinux_* *win32 *_i686 *_s390x"
+      CIBW_ARCHS_MACOS: native
```
jakirkham, line 33: Think we want to make sure we are building for `x86_64` (native I believe) and `arm64` (for M1,  M2, etc.). Both we already build for. So it would be good to continue supporting them

```suggestion
      CIBW_ARCHS_MACOS: "x86_64 arm64"
```

.github/workflows/wheel.yaml
```diff
@@ -18,14 +30,25 @@ jobs:
...
 
     steps:
       - uses: actions/checkout@v4
         with:
           submodules: true
 
+      - name: Restrict number of wheel builds on PRs
+        if: ${{ !env.BUILD_ALL }}
+        run: |
+          echo "CIBW_BUILD=cp310-win_amd64 cp311-manylinux_x86_64 cp312-macosx_x86_64 cp312-macosx_arm64" >> $GITHUB_ENV
```
jakirkham, line 46: Should we add Python 3.13 in this list? Perhaps using the other macOS build?
 
```suggestion
          echo "CIBW_BUILD=cp310-win_amd64 cp311-manylinux_x86_64 cp312-macosx_x86_64 cp313-macosx_arm64" >> $GITHUB_ENV
```

.github/workflows/wheel.yaml
```diff
@@ -6,9 +6,21 @@ concurrency:
...
+  # Build all wheels on either:
+  #    - tagged release
+  #    - pull request opened with "test all wheels" label
+  #    - pull requests when "test all wheels" label is added
+  BUILD_ALL: |
+    ${{
+        (github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags/v'))
+        || contains(github.event.pull_request.labels.*.name, 'test all wheels')
+        || (github.event.action == 'labeled' && github.event.label.name == 'test all wheels')
+    }}
```
jakirkham, line 19: Ok so in this case we are relying on maintainers to add the label for the wheel builds

Is it worth considering an option users can opt-in with? Like commit message text or PR title?

jakirkham (commented):
Thanks David! 🙏

Agree this is a good idea. Had a few questions below

jakirkham:
Any thoughts on my comments above?

dstansby:
Sorry, I've been distracted by other stuff - thanks for the comments, will try and get round to this soon!

dstansby:
Going to close, see https://github.com/zarr-developers/numcodecs/issues/288#issuecomment-2597676787

PR #565, opened 2024-08-24
Fix pre-commit

dstansby:
This makes pre-commit pass again. Not sure how this got through...


DimitriPapadopoulos:
Starting with ruff 0.6, notebooks are linted as well. This was not the case before.

Issue #464, opened 2023-09-25
Support external lz4, zstd, blosc


haampie:
#### Problem description

This package seems to unconditionally build vendored copies of lz4, zstd and blosc. This (a) makes it hard to integrate this package with other libraries that pull in any of these libraries at a different version, and (b) update binary dependencies in case there are CVEs, and (c) deal with patches of those binary dependencies.

Can you please make this package work with an external/system lz4, zstd, and blosc? That'd be very helpful, thanks.

martindurant:
I have mentioned elsehwere, but the package `cramjam` includes several useful and important compressors in one install, and ther tend to have good preformance compared to other versions. It does *not* support blosc, but that could be requested and would make numcodecs a much much simpler package to build and distribute.

joshmoore:
Hi @haampie. I don't think anyone will disagree with your thesis here. It's more a matter of making it happen. Helping hands and suggestions like @martindurant's on getting us there are very welcome!

BwL1289:
Linking #254. I know this is quite old. I'm experiencing this as well.

martindurant:
Absolutely, thanks @milesgranger . The code here still references blosc1, so some compat work would be needed I think.

I really think that there is no need for numcodecs to be repeating the build process for these standard compressors and/or relying on system libraries. Some things will still need cython for custom algorithms, but the less the better.

Issue #314, opened 2022-03-25
`zstd` dependency?


TheChymera:
I see that the code mentions zstd quite a bit, but does not list it in the dependences.
Is this an ommission, or does the package want to use its own zstd?
If so, is there any way to tell it to use the system zstd distribution?

```
chymera@decohost ~/src/numcodecs $ ag zstd -o | wc -l
703
chymera@decohost ~/src/numcodecs $ ag zstd requirements* -o | wc -l
0
```

joshmoore:
Hi @TheChymera. zstd is being pulled in via the `c-blosc/` submodule. I'll defer to others whether or not there could be a way to use a different version.

martindurant:
I'd like to note the package [`cramjam`](https://github.com/milesgranger/pyrus-cramjam), which includes several de/compression algorithms optimised for numpy buffers and in a single small static binary. It does *not* blosc, so no blosc frame headers. This is what fastparquet uses.

Issue #264, opened 2020-12-28
How to unbundle blosc, zstd, lz4


bnavigator:
From a distribution point of view, it is undesirable to bundle outdated standard system libraries into packages. Here is a quick guide how to unbundle blosc, zstd and lz4 from numcodecs but use the system libraries instead:

1. delete c-blosc subdirectory
2. Patch setup.py to link the extensions against the libraries:
```patch
Index: numcodecs-0.7.2/setup.py
===================================================================
--- numcodecs-0.7.2.orig/setup.py
+++ numcodecs-0.7.2/setup.py
@@ -112,6 +112,7 @@ def blosc_extension():
         Extension('numcodecs.blosc',
                   sources=sources + blosc_sources,
                   include_dirs=include_dirs,
+                  libraries=[] if blosc_sources else ['blosc'],
                   define_macros=define_macros,
                   extra_compile_args=extra_compile_args,
                   ),
@@ -152,6 +153,7 @@ def zstd_extension():
         Extension('numcodecs.zstd',
                   sources=sources + zstd_sources,
                   include_dirs=include_dirs,
+                  libraries=[] if zstd_sources else ['zstd'],
                   define_macros=define_macros,
                   extra_compile_args=extra_compile_args,
                   ),
@@ -185,6 +187,7 @@ def lz4_extension():
         Extension('numcodecs.lz4',
                   sources=sources + lz4_sources,
                   include_dirs=include_dirs,
+                  libraries=[] if lz4_sources else ['lz4'],
                   define_macros=define_macros,
                   extra_compile_args=extra_compile_args,
                   ),

```

The modification to setup.py would need some more work, but maybe you want to consider adding the possibility to select the building against system libraries through some environment variable, e.g. `USE_SYSTEM_LIBS=1`.

This also resolves the problems reported in #215

jakirkham:
Yeah definitely understand the value of unbundling. However this was impractical until recently as there were not Blosc wheels. As that has now changed ( https://github.com/zarr-developers/numcodecs/issues/262 ), agree it makes sense to use external dependencies.

To solve this would actually go about this a bit differently. In particular would change codecs to use 3rd party libraries in the code and drop the Blosc submodule.

jakirkham:
Just to add PR ( https://github.com/zarr-developers/numcodecs/pull/274 ) started down this path

hmaarrfk:
[...]

Issue #424, opened 2023-03-17
zstd failure for unsepcified buffer size


martindurant:
Reported in kerchunk: https://github.com/fsspec/kerchunk/issues/317#issuecomment-1470970569 by @cgohlke

#### Minimal, reproducible code sample, a copy-pastable example if possible

```python
>>> import numcodecs, cramjam
>>> numcodecs.Zstd().decode(cramjam.zstd.compress(bytes(1)))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "numcodecs/zstd.pyx", line 219, in numcodecs.zstd.Zstd.decode
  File "numcodecs/zstd.pyx", line 153, in numcodecs.zstd.decompress
RuntimeError: Zstd decompression error: invalid input data
```

#### Problem description

> According to [zstd.h, note 2](https://github.com/facebook/zstd/blob/e2965edd107acd0123294e23848cdc00af906468/lib/zstd.h#L168-L190) "decompressed size is an optional field, it may not be present, typically in streaming mode."

This means, that we can always decompress chunks written by our own Zstd, but. not necessarily those written by others. This is another case in which we could simply use cramjam or imagecodecs (or even zstandard) rather than building out own cython extension. In most cases, we will know the size of the decompressed data, because it must fit the output (but other filters in the chain may spoil this).

#### Version and installation information

Please provide the following:

* version: main
* Version of Python interpreter: any

ilan-gold:
FWIW I have now run into this twice with other `zstd` implementations.  Here's a quick reproducer of one

```
uv venv venv
source venv/bin/activate
uv pip install zarrs zarr numpy
```
```ipython
import zarr
import zarrs
import numpy as np

default_pipeline = zarr.config.get("codec_pipeline.path")
with zarr.config.set({"codec_pipeline.path": "zarrs.ZarrsCodecPipeline"}):

    memory_array = np.array(25)
    z = zarr.create_array("foo.zarr", shape=memory_array.shape, dtype=memory_array.dtype, compressors=[zarr.codecs.ZstdCodec()])
    z[...] = memory_array

with zarr.config.set({"codec_pipeline.path": default_pipeline}):
    z = zarr.open_array("foo.zarr")
    z[...]
````

The other one is the GPU implenetation from nvcomp 3.0 as implemented/tested in https://github.com/ilan-gold/cuda-zarr

Env:
<details>
```
asciitree==0.3.3
asttokens==3.0.0
[...]

PR #569, opened 2024-08-27
Use System lz4, zstd, and blosc

BwL1289:
This PR adds the capability to build Python bindings against system-provided `blosc`, `zstd`, `lz4`, and blosc linked `zlib-ng` instead of the vendored libraries included in `numcodecs`.

This PR addresses #464, #264, and by extension the comment in [#314](https://github.com/zarr-developers/numcodecs/issues/314#issuecomment-1079336948).


BwL1289:
@martindurant bumping this if it's not too inconvenient for you. Let me know how I can help support to get it merged.

martindurant:
I may not be the one you want to review this one. Actually I don't understand: does this build wheels which can then be used without the given system libraries?

Previously, I have been trying to argue, yes, that vendoring is unnecessary, but because we can get the same libraries via conda or wheels (e.g., cramjam, which now supports blosc2 also, all in static libs; [here](https://github.com/milesgranger/cramjam/issues/95) is a discussion on zlib-ng).

BwL1289:
This allows to build numcodecs against already installed system libraries instead of the vendored ones. 

Using vendored libs when you're compiling & using those libraries yourself in other places may lead to strange and hard to debug bugs due to binary incompatibilities as not only API and ABI have to be the same in such cases and objects must be fully identical in RAM which includes hidden fields (which are affected by defines).

Looping in @joshmoore if that's OK. 

Thanks so much.

martindurant:
By "statically inked", I mean this is what cramjam does internally, so for numcodecs it would be a simple dependency, no compile step at all.

martindurant:
I don't know, and I can't say for sure it will happen - I just think it should. This is further complicate by the changes required by zarr3.

joshmoore:
Thanks, @BwL1289. :+1: for having this as a new flag that leaves everything untouched unless someone invokes it. The only possible downside I can imagine is the (slight) increase in complexity of the build logic, but it looks generally well done. From my side, no objections. Pinging @jakirkham and @dstansby for any potential concerns.

(This should also not dissuade anyone from picking up the [cramjam](https://github.com/milesgranger/cramjam) flag, of course :smile:)

BwL1289:
Thank you @joshmoore, much appreciated. 

And agreed re: cramjam.

I think the slight increase in build complexity is a worthwhile tradeoff in the interim for folks who need the functionality.

Happy to make improvements wherever possible.

dstansby:
Could you expand a bit on what the use case is of this? I think all the use cases described in https://github.com/zarr-developers/numcodecs/issues/464 would be fixed by depending on Python packages such that we don't need to build the libraries ourselves.

On a project level I'm reluctant about this because it removes the tight coupling between numcodecs versions and compressor library versions, and opens the possibility of numcodecs binaries existing in the wild that have arbiratry versions of compression libraries included, which then makes it harder for us to debug what loreoks like a 'standard' numcodecs package.

On a practical level I'm reluctant about this because it is untested, and adds yet more custom code to setup.py to maintain (and we are short on maintenance time!)... the second worry would be much less of a worry if this was tested, so I think a condition to merging this should be having a test run in GitHub actions that checks the new code in setup.py works.

FWIW I've also been playing around with un-bundling sources, starting with blosc: https://github.com/zarr-developers/numcodecs/pull/569. I've just opened that as a proof of concept draft, but if there's demand for that work to continue then I'm happy to push it forward (I guess using cranjam + blosc as the two external libraries numcodecs depends on?).

QuLogic:
FWIW, I've been running with [unbundled libraries on Fedora](https://github.com/zarr-developers/numcodecs/compare/main...QuLogic:numcodecs:fedora-patches) for several years, and have not had any failures in tests. Being able to do this without patches would be welcome, but I can continue to rebase those patches for my own use.

As for custom code in `setup.py`, I know it's a big change, but the move to Meson in Matplotlib made this sort of thing easier, as distributed wheels would use bundled subprojects, while distributors could disable them and link against existing libraries.

But of course, moving to not needing the external libraries would be even simpler.

doronbehar:
> On a practical level I'm reluctant about this because it is untested, and adds yet more custom code to setup.py to maintain (and we are short on maintenance time!)... the second worry would be much less of a worry if this was tested, so I think a condition to merging this should be having a test run in GitHub actions that checks the new code in setup.py works.

Indeed a GitHub action would have been great. I can say as a downstream maintainer of `numcodecs` on NixOS that I'd love to see this get merged, as it saves disk space for users of our distribution and opens many other customization possibilities for us. Even without a GitHub action, we should be able to report to you issues with this patch.

<details><summary>On 0.13.1 this is what works for us:</summary>

```patch
diff --git c/setup.py w/setup.py
index b6db079..d468b10 100644
--- c/setup.py
+++ w/setup.py
@@ -9,6 +9,7 @@ import cpuinfo
 from Cython.Distutils.build_ext import new_build_ext as build_ext
 from setuptools import Extension, setup
 from setuptools.errors import CCompilerError, ExecError, PlatformError
+import pkgconfig
 
 # determine CPU support for SSE2 and AVX2
 cpu_info = cpuinfo.get_cpu_info()
@@ -17,6 +18,7 @@ have_sse2 = 'sse2' in flags
 have_avx2 = 'avx2' in flags
 disable_sse2 = 'DISABLE_NUMCODECS_SSE2' in os.environ
 disable_avx2 = 'DISABLE_NUMCODECS_AVX2' in os.environ
+use_system_libraries = 'NUMCODECS_USE_SYSTEM_LIBS' in os.environ
 
 # setup common compile arguments
 have_cflags = 'CFLAGS' in os.environ
@@ -49,8 +51,8 @@ def error(*msg):
     print('[numcodecs]', *msg, **kwargs)
 
 
-def blosc_extension():
-    info('setting up Blosc extension')
+def _blosc_extension_with_vendored_libs():
+    info('setting up Blosc extension from vendored sources')
 
     extra_compile_args = base_compile_args.copy()
     define_macros = []
@@ -125,8 +127,68 @@ def blosc_extension():
     return extensions
 
 
-def zstd_extension():
-    info('setting up Zstandard extension')
+def _blosc_extension_with_system_libs():
+    info('setting up Blosc extension with system libraries')
+
+    extra_compile_args = base_compile_args.copy()
+
+    blosc_package_configuration = pkgconfig.parse("blosc")
+
+    define_macros = blosc_package_configuration["define_macros"]
+    include_dirs = blosc_package_configuration["include_dirs"]
+    libraries = blosc_package_configuration["libraries"]
+    library_dirs = blosc_package_configuration["library_dirs"]
+
+    # remove minizip because Python.h 3.8 tries to include crypt.h
+    include_dirs = [d for d in include_dirs if 'minizip' not in d]
+
+    # define_macros += [('CYTHON_TRACE', '1')]
+
+    # SSE2
+    if have_sse2 and not disable_sse2:
+        info('compiling Blosc extension with SSE2 support')
+        extra_compile_args.append('-DSHUFFLE_SSE2_ENABLED')
+        if os.name == 'nt':
+            define_macros += [('__SSE2__', 1)]
+    else:
+        info('compiling Blosc extension without SSE2 support')
+
+    # AVX2
+    if have_avx2 and not disable_avx2:
+        info('compiling Blosc extension with AVX2 support')
+        extra_compile_args.append('-DSHUFFLE_AVX2_ENABLED')
+        if os.name == 'nt':
+            define_macros += [('__AVX2__', 1)]
+    else:
+        info('compiling Blosc extension without AVX2 support')
+
+    sources = ['numcodecs/blosc.pyx']
+
+    # define extension module
+    extensions = [
+        Extension(
+            'numcodecs.blosc',
+            sources=sources,
+            include_dirs=include_dirs,
+            define_macros=define_macros,
+            extra_compile_args=extra_compile_args,
+            libraries=libraries,
+            library_dirs=library_dirs,
+        ),
+    ]
+
+    return extensions
+
+
+def blosc_extension():
+    if use_system_libraries:
+        return _blosc_extension_with_system_libs()
+    else:
+        return _blosc_extension_with_vendored_libs()
+
+
+def _zstd_extension_with_vendored_sources():
+    info('setting up Zstandard extension from vendored sources')
 
     zstd_sources = []
     extra_compile_args = base_compile_args.copy()
@@ -167,8 +229,46 @@ def zstd_extension():
     return extensions
 
 
-def lz4_extension():
-    info('setting up LZ4 extension')
+def _zstd_extension_with_system_libs():
+    info('setting up Zstandard extension with system libraries')
+
+    extra_compile_args = base_compile_args.copy()
+
+    zstd_package_configuration = pkgconfig.parse("libzstd")
+    include_dirs = zstd_package_configuration["include_dirs"]
+    define_macros = zstd_package_configuration["define_macros"]
+    libraries = zstd_package_configuration["libraries"]
+    library_dirs = zstd_package_configuration["library_dirs"]
+
+    # define_macros += [('CYTHON_TRACE', '1')]
+
+    sources = ['numcodecs/zstd.pyx']
+
+    # define extension module
+    extensions = [
+        Extension(
+            'numcodecs.zstd',
+            sources=sources,
+            include_dirs=include_dirs,
+            define_macros=define_macros,
+            extra_compile_args=extra_compile_args,
+            libraries=libraries,
+            library_dirs=library_dirs,
+        ),
+    ]
+
+    return extensions
+
+
+def zstd_extension():
+    if use_system_libraries:
+        return _zstd_extension_with_system_libs()
+    else:
+        return _zstd_extension_with_vendored_sources()
+
+
+def _lz4_extension_with_vendored_sources():
+    info('setting up LZ4 extension from vendored sources')
 
     extra_compile_args = base_compile_args.copy()
     define_macros = []
@@ -195,6 +295,45 @@ def lz4_extension():
     return extensions
 
 
+def _lz4_extension_with_system_libs():
+    info('setting up LZ4 extension with system libraries')
+
+    extra_compile_args = base_compile_args.copy()
+
+    lz4_package_configuration = pkgconfig.parse("liblz4")
+    include_dirs = lz4_package_configuration["include_dirs"]
+    define_macros = lz4_package_configuration["define_macros"]
+    libraries = lz4_package_configuration["libraries"]
+    library_dirs = lz4_package_configuration["library_dirs"]
+
+    include_dirs += ['numcodecs']
+    # define_macros += [('CYTHON_TRACE', '1')]
+
+    sources = ['numcodecs/lz4.pyx']
+
+    # define extension module
+    extensions = [
+        Extension(
+            'numcodecs.lz4',
+            sources=sources,
+            include_dirs=include_dirs,
+            define_macros=define_macros,
+            extra_compile_args=extra_compile_args,
+            libraries=libraries,
+            library_dirs=library_dirs,
+        ),
+    ]
+
+    return extensions
+
+
+def lz4_extension():
+    if use_system_libraries:
+        return _lz4_extension_with_system_libs()
+    else:
+        return _lz4_extension_with_vendored_sources()
+
+
 def vlen_extension():
     info('setting up vlen extension')
     import numpy
```

</details>

mschubert:
Being able to link system libs would also help distribution in [Arch](https://aur.archlinux.org/packages/python-numcodecs) and [Gentoo](https://github.com/gentoo/sci/pull/1289) (and, likely, any other distribution).

BwL1289:
I apologize for the delay. Glad to see there's a lot of interest in this getting merged. We have been using our fork for months with no issues.

Unfortunately I don't have the bandwidth to add tests (valid objection to this getting merged) unless someone is willing/able to sponsor. Any help from the community is appreciated.

Let me know!

BwL1289:
@joshmoore thanks, will merge when I can.

dstansby:
> cf [#424 (comment)](https://github.com/zarr-developers/numcodecs/issues/424#issuecomment-2681612618) :
> 
> > I'm very heavily in favour of depending on external packages instead of vendoring
> 
> (this specifically about zstd)
> 
> Clearly the two conversations are going in opposite directions. cc @dstansby

For me there's a difference between allowing the use of system packages, and using 3rd party Python libraries that are pinned to specific versions (what I'm suggesting in https://github.com/zarr-developers/numcodecs/issues/424#issuecomment-2681612618 - I neglected to mention pinning to specific versions).

There's also a difference in maintenance burden. This PR adds more maintenance burden, because it adds more code to `setup.py`. Replacing our bundled libraries with 3rd party libraries reduces maintenance burden, because we no longer have to custom compile bundled libraries.

Taking a step back, I think there is a tension between whether numcodecs is primarily:

1. A convenience package for `zarr-python`, that first and foremost serves `zarr-python`
2. A general purpose "here's a lot of codecs with the same interface" package that is first and foremost intended to be used by anyone

If it's 2., then I think this PR fits with the scope and vision. If it's 1., then I think not so much, because (in my view) as a data writer I think we as Zarr developers should be very careful about the libraries we're using to compress data, to make sure they are working as intended. Allowing linking against system libraries loosens the control we have over the code that's compressing data.

My view is biased here - I've never packaged for systems. I won't block this going in, but I also don't personally think it's a valuable direction to go in for `numcodecs` given the scarcity of maintenance time at the moment.

d-v-b:
how do you recommend testing these changes?

BwL1289:
@d-v-b manually install the required native dependencies on your system and then compile numcodecs from source using the environment variable: `NUMCODECS_USE_SYSTEM_LIBS=1`

d-v-b:
And how would you recommend doing that process as part of our automated tests?

d-v-b:
the context for my question is that this PR adds a lot of build logic, implemented in python. This needs to be testable in our automated test suite.

d-v-b:
> > now that this project is using pixi, it's pretty easy to declare compiled dependencies
> 
> These are not _system_ libraries, though. Conda has been around for a long time...

I'm not sure I follow. If we use pixi to declare dependencies sourced from conda forge, like [lz4](https://anaconda.org/conda-forge/lz4-c/), [zstd](https://anaconda.org/conda-forge/zstd), and [blosc](https://anaconda.org/conda-forge/blosc), then presumably we can compile numcodecs against the libraries included with those dependencies, and this would work as a test for the functionality in this PR. Obviously we can also do that without pixi, but pixi makes this simpler to declare as part of the regular developer workflow.

martindurant:
I probably shouldn't have written - please go ahead with your approach :) 

I suppose pixi just hasn't fit in with my workflow styles, so I've missed the point - you could as easily have added an `apt` line to the workflow or the same deps into a traditional conda environment.yaml file. They all seem equivalent to me.

mkitti:
I would very much like to see the numcodecs broken up into smaller packages each implementing a single compressor. This should also allow for different implementations of the same codec to be loaded.

We have done this in Julia. For example, the Zstandard implementation is called ChunkCodecsLibZstd.jl:

https://github.com/JuliaIO/ChunkCodecs.jl/tree/main/LibZstd

This includes a mechanism to use system binaries via an override configuration:
https://docs.binarybuilder.org/stable/jll/#Overriding-specific-products

PR #570, opened 2024-08-28
Apply normalization consistently in VLenBytes

folded:
None and 0 are treated like a 0 length string when computing lengths, and the same normalization should be applied to the value passed to PyBytes_AS_STRING. If this is not done, an assertion is hit in the python runtime (when compiled in debug mode).

TODO:

- [ ] Tests pass locally
- [ ] Docs build locally

### DIFF
--- a/numcodecs/vlen.pyx
+++ b/numcodecs/vlen.pyx
@@ -250,7 +250,10 @@ class VLenBytes(Codec):
             l = lengths[i]
             store_le32(<uint8_t*>data, l)
             data += 4
-            encv = PyBytes_AS_STRING(values[i])
+            b = values[i]
+            if b is None or b == 0:  # treat these as missing value, normalize
+                b = b''
+            encv = PyBytes_AS_STRING(b)
             memcpy(data, encv, l)
             data += l
 


numcodecs/vlen.pyx
```diff
@@ -250,7 +250,10 @@ class VLenBytes(Codec):
             l = lengths[i]
             store_le32(<uint8_t*>data, l)
             data += 4
-            encv = PyBytes_AS_STRING(values[i])
+            b = values[i]
+            if b is None or b == 0:  # treat these as missing value, normalize
```
martindurant, line 254: Like on line 231? It seems like `l = 0` in this case, there is nothing to copy, and these operations are wasted. I might be wrong...

folded: If you don't pass a bytes object, you will hit this assert when calling PyBytes_AS_STRING: https://github.com/python/cpython/blob/40fff90ae3d46843bb9d27c6a53ef61c861a3bb4/Include/cpython/bytesobject.h#L21

it would be equivalent to skip the PyBytes_AS_STRING and memcpy entirely if l = 0. Shall I update the PR to do that instead?

PR #571, opened 2024-08-31
Proof of concept for depending on external blosc python package

dstansby:
This is a proof of concept for depending on the `blosc` Python package for blosc compression, instead of using the sources we build ourselves. The goal here is to proove it's possible in as few as lines as possible. For tests I copied the existing blosc tests, and skipped the ones that test API not provided by `blosc`.

If we go ahead with this we would need to first deprecate the following functions in `numcodecs.blosc` before the replacement could be done:

```python
[
'destroy',
'init',
'compname_to_compcode',
'cbuffer_sizes',
'cbuffer_metainfo',
'err_bad_cname',
'decompress_partial'
]
```

TODO:

- [ ] Tests pass locally
- [ ] Docs build locally

### DIFF
--- a/numcodecs/blosc_v2.py
+++ b/numcodecs/blosc_v2.py
@@ -1,0 +1,180 @@
+"""
+An attempt at replacing bundled versin of blosc with
+the blosc-python package, which provides pre-build wheels.
+
+List of functions to deprecate:
+[
+'destroy',
+'init',
+'compname_to_compcode',
+'cbuffer_sizes',
+'cbuffer_metainfo',
+'err_bad_cname',
+'decompress_partial'
+]
+
+List of behaviour to deprecate:
+- Passing cname as bytes
+
+"""
+
+from numcodecs.abc import Codec
+import numpy as np
+
+import blosc
+from blosc import (
+    BITSHUFFLE,
+    SHUFFLE,
+    NOSHUFFLE,
+    MAX_BUFFERSIZE,
+    MAX_THREADS,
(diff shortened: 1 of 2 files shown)


d-v-b:
looking back at the commit history, it seems like blosc has been vendored in this repo for a very long time. @alimanfoo and / or @jakirkham what do you think about this PR?

normanrz:
Since we are using the zstd and lz4 code from within the c-blosc repo, how would we deliver those codecs when using the blosc wheels?

alimanfoo:
Hi @dstansby, this was a while ago now but IIRC part of the original motivation for vendoring was to be able to minimise memory copies. So maybe worth a check on whether using the Python `blosc` package would introduce any extra memory copies.

normanrz:
Hey @dstansby, can you remind me why it is desirable to use the external blosc package instead of the submodule? I can see a number of downsides but I am not clear about the upsides.

dstansby:
My idea here was to do a minimal proof of concept, to see if the tests still pass (ie, data is still encoded the same way) if we depend on the blosc Python package.

For me, the advantages would be:

- Remove the need to compile c-blosc every time numcodecs is installed from source
- Remove a bunch of wrapper code from numcodecs, reducing maintenance burden

I'm not really sure if this is desirable or not overall though, and from reading various PRs and issues I think there's still an open question about the future of numcodecs. Hopefully this shows that it's technically possible to switch to the Python blosc package though.

PR #584, opened 2024-09-27
Increase speed of Delta encoding using np.subtract

ehgus:
This PR accelerated Delta encoding by replacing `np.diff` into `np.subtract`.
Using inplace operation of `np.subtract` reduce array creation time.

Here is a simple benchmark:
```Python
from numcodecs import Delta
from numcodecs.compat import ensure_ndarray
import numpy as np
import time

# new codec
class CustomDelta(Delta):
    def encode(self, buf):
        arr = ensure_ndarray(buf).view(self.dtype)
        arr = arr.reshape(-1, order='A')
        enc = np.empty_like(arr, dtype=self.astype)
        enc[0] = arr[0]

        # only difference from Delta!!
        np.subtract(arr[1:], arr[0:-1], out = enc[1:])

        return enc

data_length = int(1e9)
x = np.random.randint(0,10000,(data_length,),dtype='i4')

# benchmark
base_codec = Delta(dtype='i4')
new_codec = Delta(dtype='i4')
for codec in (base_codec, new_codec):
    time_start = time.perf_counter()
    for _ in range(10):
        codec.encode(x)
    time_end = time.perf_counter()
    benchmark_time = (time_end-time_start)/10
    print(f'time of {str(codec)} = {benchmark_time:.6f}')
# time of Delta(dtype='<i4') = 1.510299
# time of CustomDelta(dtype='<i4') = 0.953429

assert np.all(base_codec.encode(x) == new_codec.encode(x)), 'They should be equal'
```

TODO:

- [x] Tests pass locally
- [ ] Docs build locally


dstansby:
Looks good to me 👍 - could you just leave a comment above the line you changed explaining that you're using `subtract` for speed, otherwise someone might come along later and switch it back 🙈

dstansby (approved):

jakirkham:
Thanks Lee and David! 🙏

This is a good improvement. Always good to avoid unneeded array creation and copies 🙂

Noticed that `bool` needs a bit of extra handling. This wasn't tested previously. Idk to what extent it is used (maybe `Delta` + `PackBits` is effective in some cases?). Still it seems like a good idea to keep that case working given it is low effort. Made a small tweak to this effect with a test in PR: https://github.com/zarr-developers/numcodecs/pull/595

bnavigator:
Breaks zarr: #653

PR #608, opened 2024-10-16
Copy in `BitRound`

slevang:
I was benchmarking different compression options and realized `BitRound` was overwriting my array in place.

```python
import shutil
import dask.array as da
from numcodecs import BitRound

def write_to_store(array, **kwargs):
    shutil.rmtree("test.zarr", ignore_errors=True)
    array.to_zarr("test.zarr", **kwargs)
    print(array.mean().compute())

array = da.random.default_rng(42).random((1,)).persist()

write_to_store(array)
write_to_store(array, filters=[BitRound(keepbits=5)])
write_to_store(array)
```

```
0.9167441575549085
0.921875
0.921875
```

Looks like there was [discussion](https://github.com/zarr-developers/numcodecs/pull/299#discussion_r772477738) on this in the original PR but nothing came of it. Making an explicit copy seems to work fine as well, not sure which is preferred?

cc @rabernat 

TODO:

- [ ] Tests pass locally
- [ ] Docs build locally

### DIFF
--- a/numcodecs/bitround.py
+++ b/numcodecs/bitround.py
@@ -59,7 +59,8 @@ class BitRound(Codec):
             return a
         if self.keepbits > bits:
             raise ValueError("Keepbits too large for given dtype")
-        b = a.astype(a_int_dtype, copy=True)
+        b = a.copy()
+        b = b.view(a_int_dtype)
         maskbits = bits - self.keepbits
         mask = (all_set >> maskbits) << maskbits
         half_quantum1 = (1 << (maskbits - 1)) - 1
--- a/numcodecs/bitround.py
+++ b/numcodecs/bitround.py
@@ -59,12 +59,12 @@ class BitRound(Codec):
             return a
         if self.keepbits > bits:
             raise ValueError("Keepbits too large for given dtype")
-        b = a.view(a_int_dtype)
+        b = a.astype(a_int_dtype, copy=True)
         maskbits = bits - self.keepbits
         mask = (all_set >> maskbits) << maskbits
         half_quantum1 = (1 << (maskbits - 1)) - 1
-        b = b + ((b >> maskbits) & 1) + half_quantum1
-        b = b & mask
+        b += ((b >> maskbits) & 1) + half_quantum1
+        b &= mask
         return b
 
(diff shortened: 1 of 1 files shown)


numcodecs/bitround.py
```diff
@@ -63,8 +63,8 @@ def encode(self, buf):
         maskbits = bits - self.keepbits
```
jakirkham, line 63:
```suggestion
        b = b.copy()
        maskbits = bits - self.keepbits
```

numcodecs/bitround.py
```diff
@@ -63,8 +63,8 @@ def encode(self, buf):
...
         mask = (all_set >> maskbits) << maskbits
         half_quantum1 = (1 << (maskbits - 1)) - 1
-        b += ((b >> maskbits) & 1) + half_quantum1
-        b &= mask
+        b = b + ((b >> maskbits) & 1) + half_quantum1
+        b = b & mask
```
jakirkham, line 67:
```suggestion
        b += ((b >> maskbits) & 1) + half_quantum1
        b &= mask
```

jakirkham (commented):
Thanks Sam! 🙏

This is a good point to follow up

Would actually recommend just copying to start and then doing everything in-place on the copy. Otherwise we are generating a lot of temporary arrays

Mention this as we are doing compression memory usage can be an issue. So we want to reduce the amount of temporary arrays we create if possible

numcodecs/bitround.py
slevang: [This](https://github.com/zarr-developers/numcodecs/pull/299#discussion_r845319562) seems clean

slevang: nevermind, this was giving weird behavior. Explicit copy it is.

jakirkham (approved):

rabernat (approved):

martindurant:
I would add that https://github.com/zarr-developers/numcodecs/pull/584 did the opposite of this: an in-place operation for the sake of speed.

jakirkham:
Release note is in PR: https://github.com/zarr-developers/numcodecs/pull/609

PR #613, opened 2024-10-20
Adds crc32c codec

### LINKED ISSUE #610: Add crc32c codec
There's a request to add `crc32c` as a codec here, so `zarr-python` v3 can use it for sharding. I'm not entirely sure on the details, so someone more familiar might want to add a more in depth comment about this. More context at https://github.com/zarr-developers/zarr-python/issues/2391

normanrz:
Adds Crc32c codec, which appends a crc32c checksum to the input buffer when encoding and validates the checksum when decoding.

Fixes #610

TODO:

- [x] Tests pass locally
- [x] Docs build locally


dstansby:
Should this live in https://github.com/zarr-developers/numcodecs/blob/main/numcodecs/checksum32.py, where we already have `CRC32`? As I understand (from [a stackoverflow post](https://stackoverflow.com/questions/26429360/crc32-vs-crc32c)) `CRC32` and `CRC32C` are basically the same algorithm, so it makes sense to put them in the same namespace.

normanrz:
I didn't look for `checksum32.py`. Thanks for telling me.
I just looked into that. A problem is that the existing checksum codecs prepend the checksum whereas zarr3 expects the checksum to be appended. Either we single out crc32c to append the checksum or we add an option to all checksum codecs. Wdyt?

normanrz:
I moved the code to checksum32.py and refactored the tests (they had no parameterization) in https://github.com/zarr-developers/numcodecs/pull/613/commits/7e78d936de24c63023c592ba18ba8c576393c8d2

docs/checksum32.rst
```diff
@@ -13,6 +13,17 @@ CRC32
     .. automethod:: from_config
 
 
+CRC32C
+------
+.. autoclass:: CRC32C
```
dstansby, line 18: 😱 I forgot we were still doing things manually in the docs, this is a good reminder for me to fix this in another PR  😄

numcodecs/checksum32.py
```diff
@@ -1,39 +1,75 @@
...
-        return ndarray_copy(arr[4:], out)
+            raise RuntimeError(
+                f"Stored and computed {self.codec_id} checksum do not match. Stored: {expect}. Computed: {checksum}."
+            )
+        return ndarray_copy(payload_view, out)
 
 
 class CRC32(Checksum32):
     codec_id = 'crc32'
     checksum = zlib.crc32
```
dstansby, line 64: Is it worth explicitly adding `location = 'start'` here?

numcodecs/checksum32.py
```diff
@@ -1,39 +1,75 @@
...
 from .compat import ensure_contiguous_ndarray, ndarray_copy
 from .jenkins import jenkins_lookup3
 
+CHECKSUM_LOCATION = Literal['start', 'end']
+
 
 class Checksum32(Codec):
     # override in sub-class
     checksum = None
+    location: CHECKSUM_LOCATION = 'start'
```
dstansby, line 18: Would be good to add a quick docstring in `__init__` to describe the new parameter and attribute.

pyproject.toml
```diff
@@ -16,6 +16,7 @@ for use in data storage and communication applications."""
 readme =  "README.rst"
 dependencies = [
     "numpy>=1.7",
+    "crc32c>=2.7",
```
dstansby, line 19: I don't have any strong feelings either way, but is it worth making this an optional extra instead of being in the main dependencies?

dstansby (approved):
Looks good to me - just a few questions, but none of them blocking.

pyproject.toml
normanrz: I made it an optional dep. Seems like making it a core dependency later is easier than the other way around.

numcodecs/checksum32.py
normanrz: I added it to the class docstring. Unfortunately not very DRY. Not sure if docstrings can be dynamic?

normanrz:
I don't see why this PR isn't ready to merge (it might be just late)? Do I need another approval?

dstansby (approved):
Maybe just needs a fresh approval?

PR #619, opened 2024-10-28
Deprecate blosc helper functions

dstansby:
As a pre-requisite for depending on pre-built `blosc` wheels (see https://github.com/zarr-developers/numcodecs/pull/571, https://github.com/zarr-developers/numcodecs/issues/262, https://github.com/zarr-developers/numcodecs/issues/464), this deprecates several functions. The idea would be to have one deprecation cycle (say numdocecs 0.14), then remove the public functions in the next release (say numcodecs 0.15).

This uses the `deprecated` library to handle the deprecations, which does add a nother dependency, but it is quite lightweight.

### DIFF
--- a/numcodecs/blosc.pyx
+++ b/numcodecs/blosc.pyx
@@ -273,7 +273,7 @@ def compress(source, char* cname, int clevel, int shuffle=SHUFFLE,
     # check valid cname early
     cname_str = cname.decode('ascii')
     if cname_str not in list_compressors():
-        err_bad_cname(cname_str)
+        _err_bad_cname(cname_str)
 
     # setup source buffer
     source_buffer = Buffer(source, PyBUF_ANY_CONTIGUOUS)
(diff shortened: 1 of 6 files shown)


numcodecs/blosc.pyx
```diff
@@ -86,24 +87,31 @@ except ImportError:
 _importer_pid = os.getpid()
 
 
-def init():
+def _init():
     """Initialize the Blosc library environment."""
     blosc_init()
 
+init = deprecated(_init)
```
normanrz, line 94: I guess decorators don't work in .pyx files?

dstansby: They probably do, but the idea here was to leave a non-deprecated version of the functions (starting with `_`) that we can call internally, and a deprecated version which has the same name as previous.

dstansby:
Last call for comments - if no-one objects, I'll merge this in a couple of weeks.

normanrz (approved):

PR #620, opened 2024-10-29
Error on warnings when testing

dstansby:
Before https://github.com/zarr-developers/numcodecs/pull/619, it would be good to convert test warnings to errors, so we can make sure there aren't any warnings being raised that we're not expecting.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -77,7 +77,7 @@ jobs:
         shell: "bash -l {0}"
         run: |
           conda activate env
-          python -m pip install "zfpy>=1" "numpy<2"
+          python -m pip install -v ".[zfpy]"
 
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yaml
```diff
@@ -77,7 +77,7 @@ jobs:
         shell: "bash -l {0}"
         run: |
           conda activate env
-          python -m pip install "zfpy>=1" "numpy<2"
```
dstansby, line 80: This is needed to avoid `zfpy` trying to install a very old version of numpy. Instead, by installing the extra, it respects the minimum version of numpy supported by numcodecs.

dstansby:
pre-commit.ci autofix

normanrz (approved):

PR #621, opened 2024-10-29
Remove trailing .0 from CI Python version

dstansby:
Otherwise we'll need to keep manually updating the micro version!


jakirkham (approved):
Thanks David! 🙏

PR #622, opened 2024-10-29
Bump minimum numpy and Python versions

dstansby:
Time for some more version bumps to follow along with [SPEC-0](https://scientific-python.org/specs/spec-0000/#support-window)!


jakirkham:
Thanks David! 🙏

Could we also update `build-system.requires` to use `numpy>=2`?

PR #628, opened 2024-11-09
Fix minimum numpy pin

dstansby:
Fixes a mistake in https://github.com/zarr-developers/numcodecs/pull/622

Also fixes issues introduced by https://github.com/zarr-developers/numcodecs/pull/524. The problem was `zarr==3` has a `numpy>=1.25` pin, which doesn't adhere to SPEC 0 and our pin (currently `numpy>=1.24`), so on Python 3.11 don't install and run the zarr v3 tests.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -72,6 +72,9 @@ jobs:
 
       - name: Install zarr-python
         shell: "bash -l {0}"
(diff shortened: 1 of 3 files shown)


normanrz (approved):

normanrz:
Thanks!

PR #633, opened 2024-11-13
Ignore some lines for coverage

dstansby:
We (well, I) have been a bit sloppy with code coverage recently. This ignores a couple of lines that aren't covered in tests, but don't need to be.

### DIFF
--- a/numcodecs/checksum32.py
+++ b/numcodecs/checksum32.py
@@ -9,8 +9,8 @@ from .abc import Codec
 from .compat import ensure_contiguous_ndarray, ndarray_copy
 from .jenkins import jenkins_lookup3
 
-if TYPE_CHECKING:
-    from typing_extensions import Buffer  # pragma: no cover
+if TYPE_CHECKING:  # pragma: no cover
+    from typing_extensions import Buffer
 
 CHECKSUM_LOCATION = Literal['start', 'end']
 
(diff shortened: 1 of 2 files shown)


numcodecs/checksum32.py
```diff
@@ -10,7 +10,7 @@
 from .jenkins import jenkins_lookup3
 
 if TYPE_CHECKING:
```
jakirkham, line 12: Should we move this to the branch itself?

```suggestion
if TYPE_CHECKING:  # pragma: no cover
```

numcodecs/checksum32.py
```diff
@@ -10,7 +10,7 @@
...
-    from typing_extensions import Buffer
+    from typing_extensions import Buffer  # pragma: no cover
```
jakirkham, line 13:
```suggestion
    from typing_extensions import Buffer
```

numcodecs/tests/test_zarr3.py
```diff
@@ -8,7 +8,7 @@
 if not TYPE_CHECKING:
     zarr = pytest.importorskip("zarr")
 else:
```
jakirkham, line 10: Would it make sense to remove the `not` here to make this a bit clearer?

Also same question about moving the coverage pragma to the branch

```suggestion
if TYPE_CHECKING:  # pragma: no cover
    import zarr
else:
    zarr = pytest.importorskip("zarr")
```

numcodecs/tests/test_zarr3.py
```diff
@@ -8,7 +8,7 @@
...
-    import zarr
+    import zarr  # pragma: no cover
```
jakirkham, line 11:
```suggestion
```

jakirkham (commented):
Thanks David! 🙏

Had a couple questions below

jakirkham:
Thanks David! 🙏

PR #634, opened 2024-11-13
Make contributing doc page title shorter

dstansby:
Should make it look nicer in the doc header - currently it's a bit cramped:

Screenshot 2024-11-13 at 12 35 05


jakirkham:
Less is more :)

PR #637, opened 2024-11-14
Cleanup `crc32c` soft dependency

jakirkham:
Instead of creating a wrapper function around the call to `crc32c`'s to check if it can be `import`ed. Try `import`ing `crc32c` when `numcodecs.checksum32` is `import`ed. If it cannot be `import`ed, don't define the `CRC32C` class and skip registering it at the top-level with other `Codec`s.

This matches the behavior that Numcodecs has with all other optional codecs and cuts out unneeded overhead when using this checksum routine.

xref: https://github.com/zarr-developers/numcodecs/pull/613

<hr>

TODO:

- [ ] Tests pass locally
- [ ] Docs build locally

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -31,4 +31,4 @@ repos:
       hooks:
       -   id: mypy
           args: [--config-file, pyproject.toml]
-          additional_dependencies: [numpy, pytest, zfpy, 'zarr==3.0.0b1']
+          additional_dependencies: [numpy, pytest, crc32c, zfpy, 'zarr==3.0.0b1']
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -11,6 +11,11 @@ Release notes
 Unreleased
 ----------
 
+Fixes
+~~~~~
+* Cleanup ``crc32c`` soft dependency.
+  By :user:`John Kirkham <jakirkham>`, :issue:`637`
+
 
 .. _release_0.14.0:
 
--- a/numcodecs/checksum32.py
+++ b/numcodecs/checksum32.py
@@ -1,13 +1,19 @@
 import struct
 import zlib
 from collections.abc import Callable
-from typing import TYPE_CHECKING, Literal
+from contextlib import suppress
+from types import ModuleType
+from typing import TYPE_CHECKING, Literal, Optional
 
 import numpy as np
 
 from .abc import Codec
 from .compat import ensure_contiguous_ndarray, ndarray_copy
 from .jenkins import jenkins_lookup3
+
+_crc32c: Optional[ModuleType] = None
+with suppress(ImportError):
+    import crc32c as _crc32c  # type: ignore[no-redef]
 
 if TYPE_CHECKING:
     from typing_extensions import Buffer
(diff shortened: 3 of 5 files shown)


numcodecs/checksum32.py
```diff
@@ -168,3 +152,20 @@ def decode(self, buf, out=None):
...
+
+        Parameters
+        ----------
+        location : 'start' or 'end'
+            Where to place the checksum in the buffer.
+        """
+
+        codec_id = 'crc32c'
+
+        checksum = _crc32c.crc32c
```
jakirkham, line 170: So this works when simply installing `crc32c` and doing `import crc32c as _crc32c` as above

However mypy seems to [dislike this]( https://results.pre-commit.ci/run/github/68207196/1731548166.F8fD3_m4TDKH1jxKLw6gsA )

```python
numcodecs/checksum32.py:170: error: Item "None" of Module | None has no attribute "crc32c"  [union-attr]
```

.pre-commit-config.yaml
```diff
@@ -31,4 +31,4 @@ repos:
       hooks:
       -   id: mypy
           args: [--config-file, pyproject.toml]
-          additional_dependencies: [numpy, pytest, zfpy, 'zarr==3.0.0b1']
+          additional_dependencies: [numpy, pytest, crc32c, zfpy, 'zarr==3.0.0b1']
```
jakirkham, line 34: Would have thought adding `crc32c` here would be sufficient to fix the issue, but it keeps coming up

Am curious what we need to do to tell mypy it is ok to optionally use `crc32c`

jakirkham (commented):
Seeing some issues with mypy below. Flagged relevant lines that may need changes. Please let me know if there is a good way to address this

numcodecs/checksum32.py
jakirkham: Added comments for mypy to ignore the lines it was concerned about. Seems to have addressed the issue. These are not actual problems after all. Just mypy being picky

jakirkham:
Fixed the mypy issues in recent changes

numcodecs/tests/test_checksum32.py
```diff
@@ -17,6 +14,12 @@
     check_repr,
 )
 
+has_crc32c = False
+with suppress(ImportError):
```
dstansby, line 18: Instead of siliently skipping cr32c tests, I think it's clearer if they are explicitly skipped by pytest (so it's visible and easier to debug when they are skipped). I think the easiest way to do this would be to parameterize the tests that you've had to modify, and then have a conditional `pytest.skip` call inside the tests if the parmeterized parameter is related to cr32c, and the package isn't available.

numcodecs/checksum32.py
```diff
@@ -168,3 +152,19 @@ def decode(self, buf, out=None):
             out.view("uint8")[:] = b[:-4]
             return out
         return memoryview(b[:-4])
+
+
+if _crc32c:
```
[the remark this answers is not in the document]
jakirkham: This is has been the status quo for a long time

https://github.com/zarr-developers/numcodecs/blob/29995e3bff4c13a9cdf3e30a4006193c55bf9dbb/numcodecs/lzma.py#L14-L19

https://github.com/zarr-developers/numcodecs/blob/29995e3bff4c13a9cdf3e30a4006193c55bf9dbb/numcodecs/zfpy.py#L28-L35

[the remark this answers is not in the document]
jakirkham: In terms of documentation, it appears we already install `crc32c` for ReadTheDocs

https://github.com/zarr-developers/numcodecs/blob/29995e3bff4c13a9cdf3e30a4006193c55bf9dbb/.readthedocs.yaml#L17-L25

dstansby (approved):
Looks good - just needs a changelog entry 👍

docs/release.rst
```diff
@@ -11,6 +11,11 @@ Release notes
 Unreleased
 ----------
 
+Fixes
+~~~~~
+* Cleanup ``crc32c`` soft dependency.
+  By :user:`John Kirkham <jakirkham>`, :issue:`637`
```
jakirkham, line 17: Have added a release note here

jakirkham:
Thanks David! 🙏

Please let me know if there is anything else

jakirkham:
Closing and reopening for CI

PR #638, opened 2024-11-14
Fail CI if codecov doesn't upload

dstansby:
On https://github.com/zarr-developers/numcodecs/pull/637 and other PRs, it looks like the code coverage is failing because some reports fail to upload, but this isn't failing the CI jobs so it's not obvious.

This PR changes that so CI fails if coverage does not succesfully upload to codecov.


dstansby:
Looks like we're running into https://github.com/codecov/codecov-action/issues/1643 😢

PR #639, opened 2024-11-14
Support `pcodec` v0.3

### LINKED ISSUE #623: Compatibility with pcodec 0.3
`pcodec` 0.3 was released recently, and it has breaking changes. For now I've pinned to `pcodec<0.3` in https://github.com/zarr-developers/numcodecs/pull/622, but we should probably update that to `pcodec>=0.3` and adapt numcodecs to work with the newer version.

slevang:
Compatibility for the new `delta_spec` argument and modes in `pcodec.ChunkConfig`. Maintains backwards compatibility if only `delta_encoding_order` is passed. Also adds the `paging_spec` arg.

Closes #623

TODO:

- [x] Tests pass locally
- [x] Docs build locally


slevang:
Maybe we should expand the numcodecs API to support future specs?
```python
delta_spec: Literal["auto", "none", "try_consecutive", "try_lookback"] = "auto" # ignored if <v0.3
delta_encoding_order: Optional[int] = None # ignored if >=v0.3 and spec != try_consecutive?
```
cc @rabernat @mwlon

rabernat:
> Maybe we should expand the numcodecs API to support future specs?

Could you say more about what specifically you mean here? I don't quite get the context for this comment.

slevang:
@mwlon presumably understands the underlying changes much better than I do. But the gist is pcodec [now takes](https://github.com/mwlon/pcodec/pull/242/files#diff-cf8e35874c05f273925d5afcc6e70757fbb2b9a221185215cd540b24d6b39eb3) a more flexible `DeltaSpec` object, rather than a simple int that routes to auto/consecutive modes.

This broke the numcodecs interface on the latest release, so I'm just trying to figure out the right path to support any future additions without breaking any existing numcodecs users.

slevang:
Yeah that makes a lot more sense :+1:

rabernat:
A key consideration here, and an important priority for numcodecs, is backwards compatibility. Ideally any data written by Zarr will be readable for a long time into the future. This means that breaking changes in the codec parameters, which would cause decoding of existing data to fail, should be avoided.

This could be relaxed if we had some sort of versioning system for codecs. But unfortunately we don't.

For any of the proposals above, I would ask these questions:
- Can we still read old data written with earlier versions of the codec?
- Is the change future-proof to additional future evolution of the codec parameters?

slevang:
I believe these are purely API changes and feature additions, since the `ChunkConfig` is only involved in setting up the encoding step. Presumably everything that's needed for decoding is self contained.

Would be good for @mwlon to confirm though. And I do wonder if the new modes (e.g. `DeltaSpec.try_lookback`, `ModeSpec.try_float_quant`) can be decoded by older versions of pcodec, although that's much less important.

> Is the change future-proof to additional future evolution of the codec parameters?

Seems all of the config API has evolved to these spec objects, so maybe now is a good time to make the numcodecs API match fully:
```
level: int = 8,
delta_spec: DeltaSpec | None = None,
mode_spec: ModeSpec | None = None,
paging_spec: PagingSpec | None = None,
```
and the user is responsible for passing through an explicit pcodec object if they want custom behavior.

mwlon:
> Can we still read old data written with earlier versions of the codec?

Yes! Pcodec will always be able to decode older format versions.

And I think I agree with @slevang 's last comment: I'm in favor of passing the pcodec specs straight through to keep everything simple. Presumably we can convert non-spec kwargs into these (based on type) for numcodec's API backward compatibility. @rabernat do you agree?

slevang:
> and the user is responsible for passing through an explicit pcodec object if they want custom behavior

Ah nevermind I think, because numcodecs expects the codec config to be json-serializable which is a requirement of zarr. I guess we'll have to just pile on the kwargs to support additional features.

rabernat:
> It's entirely possible Ryan and I are the only ones actually using `numcodecs.pcodec`,

Definitely not the case. We have customers with petabytes of Zarr data (now a little bit less! 🙌 ) encoded with pcodec.

slevang:
Ok new approach, keep the string style args but also support `delta_spec` and `paging_spec`. It would be possible to add support for the try mode specs now, but wasn't straightforward to integrate with the existing tests since these are dtype specific. So I skipped it.

pyproject.toml
```diff
@@ -68,7 +68,7 @@ zfpy = [
     "numpy<2.0.0",
 ]
 pcodec = [
-    "pcodec>=0.2,<0.3",
+    "pcodec>=0.3",
```
mwlon, line 71: Would this cause a problem for zarr users if Pcodec rolled out 0.4 or 1.0 and broke some part of the API used here? I don't think that's likely anytime soon, but why remove the upper bound?

numcodecs/pcodec.py
```diff
@@ -44,39 +47,69 @@ class PCodec(Codec):
...
                 mode_spec = ModeSpec.auto()
-            case 'classic':
+            case "classic":
                 mode_spec = ModeSpec.classic()
             case _:
-                raise ValueError(f"unknown value for mode_spec: {self.mode_spec}")
-        paging_spec = PagingSpec.equal_pages_up_to(self.equal_pages_up_to)
+                raise ValueError(f"mode_spec {self.mode_spec} is not supported")
+
+        if self.delta_encoding_order is not None:
```
mwlon, line 77: Here we're kinda backward compatibility with current casework handling, but for future readers it might be clearer if we handled the only backward compatibility case (delta encoding order specified but not delta spec) and then separately did a match for each delta spec (including TryConsecutive).

pyproject.toml
slevang: It would, and that's what happened with 0.3. Definitely safer, and fine with me.

On the other hand, `numcodecs` doesn't seem to have any sort of upstream test job, so it will be up to us to manually identify breaking vs non-breaking changes and manually bump the version constraint. If you thought the likelihood of any more breaking API changes like this (for `standalone` and `ChunkConfig`) was unlikely until 1.0, then we could constrain only to major version?

mwlon: I think breaking changes are unlikely to happen soon, but if/when they eventually do, they're likely to affect numcodecs, regardless of 0.4 or 1.0. So I'd suggest we keep the upper bound. Ultimately up to numcodecs maintainers though.

numcodecs/pcodec.py
```diff
@@ -44,39 +47,70 @@ class PCodec(Codec):
...
                 mode_spec = ModeSpec.classic()
             case _:
-                raise ValueError(f"unknown value for mode_spec: {self.mode_spec}")
-        paging_spec = PagingSpec.equal_pages_up_to(self.equal_pages_up_to)
+                raise ValueError(f"mode_spec {self.mode_spec} is not supported")
+
+        if self.delta_encoding_order is not None and self.delta_spec is None:
+            # backwards compat for before delta_spec was introduced
+            delta_spec = DeltaSpec.try_consecutive(self.delta_encoding_order)
+        elif self.delta_encoding_order and self.delta_spec != "try_consecutive":
```
mwlon, line 80: `elif self.delta_encoding_order is not None`, to get the 0 case as well

numcodecs/pcodec.py
```diff
@@ -44,39 +47,70 @@ class PCodec(Codec):
     def __init__(
         self,
         level: int = 8,
+        mode_spec: Literal["auto", "classic"] = "auto",
+        delta_spec: Literal["auto", "none", "try_consecutive", "try_lookback"] | None = None,
```
mwlon, line 51: I realize the difference between "none" and None could be confusing - how about we make the default "auto"? And then "auto" + delta_encoding_order is the legacy case to handle.

mwlon:
@slevang @rabernat should this be merged?

rabernat:
Thanks for the ping @mwlon!

This looks good, but I'm a little concerned by the drop in coverage. We strive for 100% coverage in Numcodecs. I realize it's tedious to test all of these different error states. @slevang do you feel comfortable with the uncovered blocks?

slevang:
Pretty sure `numcodecs/pcodec.py` and this patch are 100% covered. The previous coverage failure was on an [unrelated](https://github.com/zarr-developers/numcodecs/issues/646) part of the repo. Now the coverage action seems to have stalled completely :shrug:

numcodecs/pcodec.py
```diff
@@ -44,39 +47,70 @@ class PCodec(Codec):
     def __init__(
         self,
         level: int = 8,
+        mode_spec: Literal["auto", "classic"] = "auto",
```
dstansby, line 50:
```suggestion
        *
        mode_spec: Literal["auto", "classic"] = "auto",
```
Since you're changing the argument order here anyway, it would be a good oppurtunity to make these argumetns keyword only.

numcodecs/pcodec.py
```diff
@@ -44,39 +47,70 @@ class PCodec(Codec):
...
         self.equal_pages_up_to = equal_pages_up_to
-        self.mode_spec = mode_spec
-
-    def encode(self, buf):
-        buf = ensure_contiguous_ndarray(buf)
 
+    def _get_chunk_config(self):
         match self.mode_spec:
-            case 'auto':
+            case "auto" | None:
```
dstansby, line 70: Why `| None` here? As `mode_spec` is typed in `__init__`, it can't take `None` as a value.

numcodecs/tests/test_pcodec.py
```diff
@@ -57,14 +61,23 @@ def test_config():
 
 
 def test_invalid_config_error():
-    codec = PCodec(mode_spec='bogus')
+    for param in ["mode_spec", "delta_spec", "paging_spec"]:
```
dstansby, line 64: Might be a bit nicer to use [pytest to parametrize](https://docs.pytest.org/en/stable/example/parametrize.html) the test instead of having this for loop?

dstansby (changes requested):
Thanks a bunch for this PR, and sorry for taking so long to review.

This looks great overall. I've left some questions inline, and in addition to those please could you add a changelog that outlines:
- That pcodec 0.3 is now supported
- What this means for forward and backward comaptibility with data compressed using older versions of pcodec and numcodecs
- That the order of arguments in `PCodec.__init__` has changed

numcodecs/pcodec.py
slevang: Yeah this was my biggest hesitation on backwards compatibility when I looked at this again. Alternatively, I'm open to reverting to the previous kwarg order, but it makes the signature a bit messy. Otherwise I think keyword only is a good addition.

numcodecs/pcodec.py
slevang: Leftover from a previous implementation, thanks.

numcodecs/pcodec.py
dstansby: I think we should either:
- Break the order, and make everything keyword only
- Keep the order, and make everything *new* keyword only

Happy to be led by you on what's the best option? We're still < verison 1, so can break things between releases.

slevang: @rabernat any opinion here?

rabernat: I think breaking the order and making everything keyword only is the right way to go.

I'm much less concerned about API stability for a new and evolving codec like PCodec than I am about backwards compatibility of actual data on dask.

rabernat (approved):

rabernat:
@dstansby are you :+1: on this now?

dstansby (approved):
LGTM, thanks all!

PR #648, opened 2024-11-18
Type `Buffer`'s `flags` argument

jakirkham:
Ensure the `flags` argument in `Buffer` is correctly typed so it can be passed verbatim to `PyObject_GetBuffer`. This can also avoid needless conversions to and from Python `object` type.

TODO:

- [ ] Tests pass locally
- [ ] Docs build locally

### DIFF
--- a/numcodecs/compat_ext.pyx
+++ b/numcodecs/compat_ext.pyx
@@ -12,7 +12,7 @@ from .compat import ensure_contiguous_ndarray
 cdef class Buffer:
     """Convenience class for buffer interface."""
 
-    def __cinit__(self, obj, flags):
+    def __cinit__(self, obj, int flags):
         PyObject_GetBuffer(obj, &(self.buffer), flags)
         self.acquired = True
         self.ptr = <char *> self.buffer.buf


numcodecs/compat_ext.pyx
```diff
@@ -12,7 +12,7 @@ from .compat import ensure_contiguous_ndarray
 cdef class Buffer:
     """Convenience class for buffer interface."""
 
-    def __cinit__(self, obj, flags):
+    def __cinit__(self, obj, int flags):
         PyObject_GetBuffer(obj, &(self.buffer), flags)
```
jakirkham, line 16: The `flags` argument here is only used in this `PyObject_GetBuffer` call. The signature of this function [in Cython]( https://github.com/cython/cython/blob/3bd3503384c25e31c21e340baf3d5d4e380e5256/Cython/Includes/cpython/buffer.pxd#L34 ) or [in Python]( https://docs.python.org/3/c-api/buffer.html#c.PyObject_GetBuffer ) is

```cython
int PyObject_GetBuffer(object obj, Py_buffer *view, int flags)
```

Here `flags` is an `int` specified by [a range of possible constants]( https://github.com/cython/cython/blob/3bd3503384c25e31c21e340baf3d5d4e380e5256/Cython/Includes/cpython/buffer.pxd#L8-L29 ) to provide a specific kind of `Py_buffer` object

To make sure that `flags` can be passed through unaltered to this function call, this types the `__cinit__` arguments to match

jakirkham:
Thanks David! 🙏

PR #650, opened 2024-11-18
Changelog for 0.14.1

dstansby:
I thought it would be good to get these changes out sooner rather than later, given the changes to crc32c handling are breaking. @jakirkham I expanded on the changelog entry a bit to make it clear why the change was made, and more specifically what the change was.

### DIFF
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -6,14 +6,24 @@ Release notes
     # to document your changes. On releases it will be
     # re-indented so that it does not show up in the notes.
 
+    .. _unreleased:
+
+    Unreleased
+    ----------
+
 .. _unreleased:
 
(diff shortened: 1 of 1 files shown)


slevang:
#639 should be good to go, would be nice to get that in as well.

jakirkham:
@slevang believe the goal here is to make a simple bug fix release

We can follow up on enhancements in a subsequent release

Hope that makes sense 🙂

dstansby:
Yeah, I think we just want to get out the fix/change to `crc32c` here. I will take a look at #639 though, and we can do a new 0.15.0 release when that's merged.

jakirkham:
Thanks David! 🙏

Issue #646, opened 2024-11-17
Missing coverage?


jakirkham:
In [the recent CodeCov report]( https://app.codecov.io/gh/zarr-developers/numcodecs/commit/ad1ad675d933c43e16479b512cb00beed1ade375/tree/numcodecs ), they flag [these lines]( https://app.codecov.io/gh/zarr-developers/numcodecs/commit/ad1ad675d933c43e16479b512cb00beed1ade375/blob/numcodecs/json.py#L72 ) as uncoverged

https://github.com/zarr-developers/numcodecs/blob/ad1ad675d933c43e16479b512cb00beed1ade375/numcodecs/json.py#L68-L72

It is a little unclear why as those lines were introduced with corresponding tests some time ago in PR ( https://github.com/zarr-developers/numcodecs/pull/417 ). AFAICT recent NumPy versions still need to use that fallback. In fact that behavior is expected and defined in [NEP 34]( https://numpy.org/neps/nep-0034-infer-dtype-is-object.html )

This may need further investigation to resolve

<hr>

Additionally CodeCov flags [this line]( https://app.codecov.io/gh/zarr-developers/numcodecs/commit/ad1ad675d933c43e16479b512cb00beed1ade375/blob/numcodecs/tests/test_zarr3.py#L54 ) as uncovered.

https://github.com/zarr-developers/numcodecs/blob/ad1ad675d933c43e16479b512cb00beed1ade375/numcodecs/tests/test_zarr3.py#L51-L55

Just looking at it, would naively add a `# pragma: no cover`, but wanted to mention it here in case there is something I'm missing

jakirkham:
Thanks David! 🙏

jakirkham:
Reopening as [the JSON case]( https://github.com/zarr-developers/numcodecs/blob/ad1ad675d933c43e16479b512cb00beed1ade375/numcodecs/json.py#L68-L72 ) is [still present]( https://app.codecov.io/gh/zarr-developers/numcodecs/commit/88660debd8ec528ab8faf2f924028e63561eaa61/blob/numcodecs/json.py )

jakirkham:
This is the test that would ideally cover this case

https://github.com/zarr-developers/numcodecs/blob/88660debd8ec528ab8faf2f924028e63561eaa61/numcodecs/tests/test_json.py#L61-L85

However we coerce to a NumPy array, which bypasses the `try...except...` handling above

[...]

PR #651, opened 2024-11-18
Remove unused pytest skip

dstansby:
I originally included the `pytest.skip` because in rare cases (I think debug builds of Cpython??) docstrings can return `None` instead of their contents. I think this is rare enough that we don't need to care about it, and it's better just adding a `# type: ignore` comment instead of a `pytest.skip`.

I'm hoping this gets code coverage back up to 100% again to fix https://github.com/zarr-developers/numcodecs/issues/646


jakirkham (approved):

jakirkham:
Thanks David! 🙏

This seems reasonable

If we think we do need this condition, we could also add a `# pragma: no cover`

PR #656, opened 2024-11-22
Switch `Buffer`s to `memoryview`s & remove extra copies/allocations

jakirkham:
When this was written in the code, Python's Buffer Protocol support was inconsistent across Python versions (specifically on Python 2.7). Since Python 2.7 reached EOL and it was dropped from Numcodecs, the Python Buffer Protocol support has become more consistent.

At this stage the `memoryview` object, which Cython also supports, does all the same things that `Buffer` would do for us. Plus it is builtin to the Python standard library. It behaves similarly in a lot of ways.

Given this, switch the code over to `memoryview`s internally and drop `Buffer`.

<hr>

Additionally have pushed changes to this PR to improve overall memory usage. This eliminates some unneeded copies that occurred at the ended of some codecs. Also have eliminated some temporary allocations used in some codec pipelines by allocating output buffers earlier and changing operations to act in-place. This should eliminate some spiky behavior seen recently with codecs. 

<hr>

TODO:

- [ ] Tests pass locally
- [ ] Docs build locally


jakirkham:
Planning to merge end of week if no comments

rabernat (approved):
This seems like a great improvement. Thanks so much @jakirkham!

Unfortunately my Cython isn't really good enough to provide a useful review. But the tests pass! 😆  And we are reducing LoC and simplifying many functions. So seems like a great direction.

Someone who actually groks Cython should review this for real.

d-v-b:
@jakirkham any interest in pushing this forward?

jakirkham:
Have resolved the conflicts

Also to avoid making the diff as messy have added in `try...finally...` stubs. This makes the diff easier to read

We can clean these `try...finally...`s out in a subsequent PR and exclude the changes from `git blame` to avoid confusion

Lastly added a new entry

Please let me know if anything else is needed

numcodecs/blosc.pyx
```diff
@@ -366,30 +374,38 @@ def decompress(source, dest=None):
...
         size_t nbytes, cbytes, blocksize
 
-    # setup source buffer
-    source_buffer = Buffer(source, PyBUF_ANY_CONTIGUOUS)
-    source_ptr = source_buffer.ptr
+    # obtain source memoryview
+    source_mv = memoryview(source)
+    source_pb = PyMemoryView_GET_BUFFER(source_mv)
+    if not PyBuffer_IsContiguous(source_pb, b'A'):
+        raise BufferError("`source` must contain contiguous memory")
```
d-v-b, line 389: this same sequence is repeated many times. do you think it would make sense to put it in a function?

jakirkham: It's a fair question. Think the best we could get is dropping the contiguous check throughout (saving 2 lines)

Am ok either way

d-v-b: the highlighted text is repeated verbatim 9 times in this PR, so I think replacing it with invocations of a function like
```
def get_memory_view(source) -> tuple[memory view, whatever source pb is]:
      source_mv = memoryview(source)
    source_pb = PyMemoryView_GET_BUFFER(source_mv)
    if not PyBuffer_IsContiguous(source_pb, b'A'):
        raise BufferError("`source` must contain contiguous memory")
   return source_mv, source_pb
```

would result in a lot of deduplication. but I don't know cython, so maybe there's some aspect of that that I'm missing?

d-v-b: seems like a bit of a chore then. up to you if you deem it worthwhile

jakirkham: Am ok punting on this

numcodecs/zstd.pyx
```diff
@@ -134,16 +144,15 @@ def compress(source, int level=DEFAULT_CLEVEL, bint checksum=False):
...
 
     # check compression was successful
     if ZSTD_isError(compressed_size):
         error = ZSTD_getErrorName(compressed_size)
         raise RuntimeError('Zstd compression error: %s' % error)
 
     # resize after compression
-    dest = dest[:compressed_size]
+    dest_objptr = <PyObject*>dest
+    _PyBytes_Resize(&dest_objptr, compressed_size)
```
jakirkham, line 155: Per issue ( https://github.com/zarr-developers/numcodecs/issues/717 ), think this should fix the issue in Blosc (have pushed similar changes for the other codecs in this PR)

In particular this code had created a copy of the `bytes` object (trimmed to `compressed_size`)

```cython
dest = dest[:compressed_size]
```

To fix this we now call [`_PyBytes_Resize`]( https://docs.python.org/3/c-api/bytes.html#c._PyBytes_Resize ), which will resize the existing `bytes` object. This uses [`PyObject_Realloc` under-the-hood]( https://github.com/python/cpython/blob/898e6b395e63ad7f8bbe421adf0af8947d429925/Objects/bytesobject.c#L3248-L3249 ). A good `realloc` implementation can detect when the memory requested is less than the size of the original allocation (as is the case here). When that happens it will shrink the allocation in-place (meaning no new memory allocation and no-copy), this is a fast operation

numcodecs/blosc.pyx
jakirkham: Went ahead and added it. Maybe it is useful to have this in a standard function

numcodecs/compat_ext.pxd
```diff
@@ -1,12 +1,11 @@
...
-        size_t nbytes
-        size_t itemsize
-        bint acquired
-
-    cpdef release(self)
+cdef extern from *:
+    """
+    #define PyBytes_RESIZE(b, n) _PyBytes_Resize(&b, n)
+    """
+    int PyBytes_RESIZE(object b, Py_ssize_t n) except -1
```
jakirkham, line 8: Have added this function (macro) to help with resizing `bytes` objects in Cython

This allows us to truncate `bytes` allocations that are larger than we end up needing without needing to copy to a new `bytes` object

jakirkham, line 8: Have added this function (macro) to help with resizing `bytes` objects in Cython

This allows us to truncate `bytes` allocations that are larger than we end up needing without needing to copy to a new `bytes` object

numcodecs/lz4.pyx
```diff
@@ -97,17 +101,15 @@ def compress(source, int acceleration=DEFAULT_ACCELERATION):
...
+        pass
 
     # check compression was successful
     if compressed_size <= 0:
         raise RuntimeError('LZ4 compression error: %s' % compressed_size)
 
     # resize after compression
     compressed_size += sizeof(uint32_t)
-    dest = dest[:compressed_size]
+    PyBytes_RESIZE(dest, compressed_size)
```
jakirkham, line 112: Thus when we need to truncate a `bytes` object, we can avoid `[:<n>]`, which copies the `bytes` object

Instead we just call `PyBytes_RESIZE`, which can do an in-place size reduction of the `bytes` object

Any excess memory goes back to the memory pool, which Python can use as it sees fit

jakirkham (commented):
Added a few notes on recent changes to reduce memory usage by doing in-place size reductions of objects (instead of copies as was the case before)

Should help with spiky memory consumption that @tomwhite spotted recently: https://github.com/zarr-developers/numcodecs/issues/717

d-v-b:
@jakirkham thanks so much for this, it looks like a great improvement. I am not qualified to review this because I don't know cython, is there anyone you can recommend?

d-v-b:
i'm also happy merging after self-review

numcodecs/lz4.pyx
tomwhite: I have just tested this using the Mac 14 wheels from https://github.com/zarr-developers/numcodecs/actions/runs/14105743382 using the memray code in https://github.com/tomwhite/memray-array - and can confirm that it works. I can see the memory saving (peak memory goes form 300MB to 200MB). This was using zstd with Zarr v3.

jakirkham:
Thanks Davis! 🙏

@dstansby would you have time to look within the next week? 🙂

dstansby:
Sorry, I don't think I have enough C/Cython experience to review this.

jakirkham:
No worries. In that case will go ahead and merge

Ryan had reviewed it earlier. Also the fact that Davis took a look and Tom tested it successfully provides some additional confidence. Cython knowledge isn't as common around here. So that is likely as good as we can do

Am AFK for until the following week. Though happy to follow up on anything at that point. Please do feel free to ping me (otherwise it will likely get buried in the avalanche of notifications in the interim 😅)

numcodecs/fletcher32.pyx
```diff
@@ -63,23 +68,41 @@ class Fletcher32(Codec):
     def encode(self, buf):
         """Return buffer plus 4-byte fletcher checksum"""
         buf = ensure_contiguous_ndarray(buf).ravel().view('uint8')
-        cdef const uint8_t[::1] b_ptr = buf
-        val = _fletcher32(b_ptr)
-        return buf.tobytes() + struct.pack("<I", val)
+        cdef const uint8_t[::1] b_mv = buf
+        cdef uint8_t* b_ptr = &b_mv[0]
```
jakirkham, line 72:
```suggestion
        cdef const uint8_t[::1] b_mv = buf
        cdef const uint8_t* b_ptr = &b_mv[0]
```

numcodecs/fletcher32.pyx
```diff
@@ -63,23 +68,41 @@ class Fletcher32(Codec):
...
+        return out
 
     def decode(self, buf, out=None):
         """Check fletcher checksum, and return buffer without it"""
         b = ensure_contiguous_ndarray(buf).view('uint8')
-        cdef const uint8_t[::1] b_ptr = b[:-4]
-        val = _fletcher32(b_ptr)
-        found = b[-4:].view("<u4")[0]
+        cdef const uint8_t[::1] b_mv = b
+        cdef uint8_t* b_ptr = &b_mv[0]
```
jakirkham, line 88:
```suggestion
        cdef const uint8_t[::1] b_mv = buf
        cdef const uint8_t* b_ptr = &b_mv[0]
```

jakirkham (commented):
Noticed that two of the pointer definitions below should have been marked `const` to match the associated typed-memoryviews, from which they are taken. However they were not

As a result, the compiler generates warnings about these ( https://github.com/zarr-developers/numcodecs/issues/725 ). That said, we don't actually change the content the pointer is related to. So this is only a compile time warning, there is no actual issue at runtime

To fix the compiler warnings, added these `const`s (as shown below) in PR ( https://github.com/zarr-developers/numcodecs/pull/728 ), which fixes this issue

numcodecs/compat_ext.pxd
```diff
@@ -1,12 +1,11 @@
...
-        Py_buffer buffer
-        size_t nbytes
-        size_t itemsize
-        bint acquired
-
-    cpdef release(self)
+cdef extern from *:
+    """
+    #define PyBytes_RESIZE(b, n) _PyBytes_Resize(&b, n)
+    """
```
jakirkham, line 7: This should have an `#ifndef` guard to protect against redefinition error

So far it seems Cython doesn't encounter this error as it defines this once per module. Still it is a good idea to add the guard

Handling in PR: https://github.com/zarr-developers/numcodecs/pull/732

Issue #653, opened 2024-11-20
zarr 2.18.3 test suite fails on numcodecs 0.14 (test_array_with_delta_filter)


bnavigator:
Latest released zarr is still 2.8.13, but numcodecs is already 0.14. They do not play together:

```
[   97s] _________________________ test_array_with_delta_filter _________________________
[   97s] [gw0] linux -- Python 3.12.7 /usr/bin/python3.12
[   97s]
[   97s]     def test_array_with_delta_filter():
[   97s]         # setup
[   97s]         astype = "u1"
[   97s]         dtype = "i8"
[   97s]         filters = [Delta(astype=astype, dtype=dtype)]
[   97s]         data = np.arange(100, dtype=dtype)
[   97s]
[   97s]         for compressor in compressors:
[   97s] >           a = array(data, chunks=10, compressor=compressor, filters=filters)
[   97s]
[   97s] zarr/tests/test_filters.py:40:
[   97s] _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
[   97s] zarr/creation.py:444: in array
[   97s]     z[...] = data
[   97s] zarr/core.py:1449: in __setitem__
[   97s]     self.set_basic_selection(pure_selection, value, fields=fields)
[   97s] zarr/core.py:1545: in set_basic_selection
[   97s]     return self._set_basic_selection_nd(selection, value, fields=fields)
[   97s] zarr/core.py:1935: in _set_basic_selection_nd
[   97s]     self._set_selection(indexer, value, fields=fields)
[   97s] zarr/core.py:1988: in _set_selection
[   97s]     self._chunk_setitem(chunk_coords, chunk_selection, chunk_value, fields=fields)
[   97s] zarr/core.py:2261: in _chunk_setitem
[   97s]     self._chunk_setitem_nosync(chunk_coords, chunk_selection, value, fields=fields)
[   97s] zarr/core.py:2271: in _chunk_setitem_nosync
[   97s]     self.chunk_store[ckey] = self._encode_chunk(cdata)
[   97s] zarr/core.py:2387: in _encode_chunk
[   97s]     chunk = f.encode(chunk)
[   97s] _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
[...]

PR #595, opened 2024-10-10
Update optimized `Delta` codec for `bool`

jakirkham:
Previously `bool` just worked with `Delta`. However this was not actually tested. The optimized version switched to `np.subtract` for in-place computation, which works for other types. Though `bool` needs special handling. Fortunately this can be done with `np.not_equal`, which has the same behavior.

Also include a test for `bool` data to make sure this is handled correctly going forward.

TODO:

- [ ] Unit tests and/or doctests in docstrings
- [ ] Tests pass locally
- [ ] Docstrings and API docs for any new/modified user-facing classes and functions
- [ ] Changes documented in docs/release.rst
- [ ] Docs build locally
- [ ] GitHub Actions CI passes
- [ ] Test coverage to 100% (Codecov passes)

PR #677, opened 2025-01-04
Revert change to delta filter

dstansby:
This reverts https://github.com/zarr-developers/numcodecs/pull/584, as it has caused two issues so far, one of which is unresolved: https://github.com/zarr-developers/numcodecs/issues/653 and https://github.com/zarr-developers/numcodecs/pull/595.

There is a proposed fix at https://github.com/zarr-developers/numcodecs/pull/657, but that is held up under review.

In order to un-break zarr-python, I think reverting here is the right thing to do (so we can do a 0.15 release with other stuff).

My intention is to add a test to https://github.com/zarr-developers/numcodecs/issues/653 to this PR too, so keeping as draft. Opening so there's time for folks to see that I'm intending to do this; cc @jakirkham @ehgus

### DIFF
--- a/numcodecs/delta.py
+++ b/numcodecs/delta.py
@@ -63,12 +63,7 @@ class Delta(Codec):
         enc[0] = arr[0]
 
         # compute differences
-        # using np.subtract for in-place operations
-        if arr.dtype == bool:
-            np.not_equal(arr[1:], arr[:-1], out=enc[1:])
-        else:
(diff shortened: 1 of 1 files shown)


ehgus:
I also hope PR #657 will be reviewed soon as it fixes issue #653, which was caused by unsafe casting. In addition, this PR includes improved tests. Thank you for reminding me.

PR #680, opened 2025-01-05
Remove ndarray_like code

dstansby:
Almost all of this can be replaced with `numpy.typing`, which I'm guessing didn't exist when this was first written. This cleans out code we don't need to maintain any more, and makes it a lot easier to go ahead with typing the rest of the library.


dstansby:
Ha, turns out this is not as simple as I thought it would be, so closing.

PR #681, opened 2025-01-06
Zstd: Don't persist the checksum param if false

normanrz:
Changes the zstd codec to not persist the `checksum` param, if it is set to `False` (the default). This is aimed at improving backwards compatibility.

See https://github.com/zarr-developers/zarr-python/issues/2647#issuecomment-2571758778

- [x] Tests pass locally
- [x] Docs build locally

### DIFF
--- a/numcodecs/tests/test_zstd.py
+++ b/numcodecs/tests/test_zstd.py
@@ -90,3 +90,10 @@ def test_native_functions():
     assert Zstd.default_level() == 3
     assert Zstd.min_level() == -131072
     assert Zstd.max_level() == 22
+
+
+def test_zstd_config():
+    # Testing that checksum is removed from config if False
+    codec = Zstd(level=5, checksum=False)
+    config = codec.get_config()
+    assert config == {"id": "zstd", "level": 5}
--- a/numcodecs/zstd.pyx
+++ b/numcodecs/zstd.pyx
@@ -258,6 +258,13 @@ class Zstd(Codec):
              self.level)
         return r
 
+    # Override to remove "checksum" if False
+    def get_config(self):
+        config = {'id': self.codec_id, 'level': self.level}
+        if self.checksum:
+            config['checksum'] = True
+        return config
+
     @classmethod
     def default_level(cls):
         """Returns the default compression level of the underlying zstd library."""


normanrz:
@dstansby Do you know why the macos tests are failing? I don't have that issue on my local mac.

dstansby:
It's the same failure at https://github.com/zarr-developers/numcodecs/pull/679, so not related to either PR as far as I can see.

normanrz:
I don't understand how special-casing in the client library (zarr-python) is any less complex than this small fix?

I don't know what numcodecs' policy for backwards compatibility is, but conditionally adding parameters seems like a good way of dealing with that.

dstansby:
Yes, complexity of the fix would be the same, but `numcodecs` is used by a much wider range of packages than just `zarr-python`. Because this change introduces non-intuitive behaviour (only serialising a compression parameter if it has a particular value) that is not done for any other codecs in numcodecs, it should be down to `zarr-python` to choose a different way of serialising if it wants to.

normanrz:
> but `numcodecs` is used by a much wider range of packages than just `zarr-python`.

Is it really?

normanrz:
I suppose most of them have a transitive dependency through zarr.

Anyways. I don't care too strongly about this issue.

dstansby:
In that case shall we close this, and I can add a note to https://github.com/zarr-developers/zarr-python/issues/2647 that a fix/change to the codec serialisation should go in `zarr-python`?

normanrz:
Closing in favor of https://github.com/zarr-developers/zarr-python/pull/2655

nhz2:
Not having this change in numcodecs makes things more complicated for other Zarr v2 implementations. Ref: https://github.com/JuliaIO/Zarr.jl/issues/193. IMO, not being able to use numcodecs as a source of truth for this makes compatibility testing trickier than it has to be.

Also, since zarr v2 data with `"checksum": false` already exists in the wild (and are still produced by zarr-python 2.18.7), surely TensorStore needs to be updated to handle parsing compressor configs in a more flexible way.

nhz2:
Yes, having more explicit documentation is great.

Also, I'm confused why adding a new optional config property can be considered a breaking change.

Zarr.jl will ignore any compressor properties it doesn't know about, so it would not be broken by this kind of change.

PR #689, opened 2025-01-14
Minor UX improvement: Add the `UnknownCodecError`.

cwognum:
See https://github.com/zarr-developers/zarr-python/issues/2508 for context. 

Since `numcodecs` supports registering custom codecs in downstream libraries, there can be situations where the environment in which data was encoded is not the same as the environment in which the data is decoded (e.g. opening a Zarr archive that was created in a different environment). This can lead to missing codecs.

It would be a nice UX improvement if you catch that error and can offer codec specific solutions.
 
This change would enable something like: 

```py
import numcodecs

try: 
    numcodecs.get_codec({"codec_id": "imagecodecs_jpeg2k"})
except UnknownCodecError as error:
    if error.codec_id == "imagecodecs_jpeg2k":
        raise RuntimeError("This Zarr archive requires `imagecodecs`, to install it use ...") from error
```

TODO:

- [x] Tests pass locally
- [x] Docs build locally

### DIFF
--- a/numcodecs/errors.py
+++ b/numcodecs/errors.py
@@ -5,7 +5,7 @@ This module defines custom exceptions that are raised in the `numcodecs` codebas
 class UnknownCodecError(Exception):
     """
     An exception that is raised when trying to receive a codec that has not been registered.
-    
+
     Parameters
     ----------
     codec_id : str
--- a/numcodecs/errors.py
+++ b/numcodecs/errors.py
@@ -1,0 +1,17 @@
+"""
+This module defines custom exceptions that are raised in the `numcodecs` codebase.
+"""
+
+class UnknownCodecError(Exception):
+    """
+    An exception that is raised when trying to receive a codec that has not been registered.
+    
+    Parameters
+    ----------
+    codec_id : str
+        Codec identifier.
+    """
+
+    def __init__(self, codec_id: str):
+        self.codec_id = codec_id
+        super().__init__(f"codec not available: '{codec_id}'")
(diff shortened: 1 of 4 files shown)


numcodecs/errors.py
```diff
@@ -0,0 +1,26 @@
+"""
+This module defines custom exceptions that are raised in the `numcodecs` codebase.
+"""
+
+
+class UnknownCodecError(Exception):
```
cwognum, line 6: Should we make this inherit from `ValueError`?

dstansby (commented):
I'm very 👍 doing this, thanks for the PR. I left a suggestion for the test, and also I'm a bit on the fence about `ValueError`, so let me know your thoughts there.

numcodecs/errors.py
cwognum: Since a `ValueError` was thrown originally, I simply assumed that was a well motivated choice.

My question was rather because changing the exception type could break code downstream. For example, in Polaris [we explicitly catch the `ValueError`](https://github.com/polaris-hub/polaris/blob/6a96b94bcc23a599baa07ff031676feb64ac7393/polaris/dataset/zarr/_utils.py#L39).

d-v-b: if the error describes a codec that's a codec, but unknown, then it should be a `ValueError` IMO

dstansby: Ah, if it was originally a `ValueError`, then definitely 👍 to keep it as inheriting from `ValueError`

cwognum:
@d-v-b @dstansby Ready for another round of review!

d-v-b (approved):

PR #702, opened 2025-02-09
(chore) Type hints for abc codec

d-v-b:
I'm not sure how to type the buffer parameters. As long as we have a `pickle` codec, I don't think it's possible to restrict the return type of `encode` and `decode` to anything narrower than `object`.

- [ ] Tests pass locally
- [ ] Docs build locally

### DIFF
--- a/numcodecs/abc.py
+++ b/numcodecs/abc.py
@@ -29,14 +29,14 @@ other and vice versa.
 """
 
 from abc import ABC, abstractmethod
-from typing import Optional
+from typing import ClassVar
 
 
 class Codec(ABC):
     """Codec abstract base class."""
 
     # override in sub-class
-    codec_id: Optional[str] = None
+    codec_id: ClassVar[str]
     """Codec identifier."""
 
     @abstractmethod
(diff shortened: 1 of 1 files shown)


d-v-b:
i'm just annotating the type of the `codec_id` attribute -- it's a `ClassVar[str]`, with no default value not `str | None = None`

numcodecs/tests/test_registry.py
```diff
@@ -34,8 +34,8 @@ def test_all_classes_registered():
         if (
             inspect.isclass(obj)
             and issubclass(obj, numcodecs.abc.Codec)
+            and hasattr(obj, 'codec_id')
```
dstansby, line 37:
```suggestion
            and obj is not numcodecs.abc.Codec
```
I think is a bit clearer in intent (we don't want or need the base codec to be registered)?

dstansby (approved):
👍 , just one suggestion, feel free to take or leave and then merge

dstansby:
I think my suggested change (that I applied) revealed (sort of) a realy issue - `Checksum32` is a base class, not really a codec that can be used by itself. I've made this better by making it an abstract base class in https://github.com/zarr-developers/numcodecs/pull/711 - if/when that's merged, we can update the check here in the test to `not isinstance(codec, abc.ABC)`.

PR #704, opened 2025-02-09
(chore) use pixi for building

d-v-b:
Adds pixi for building the project. On my linux desktop, everything compiles with 
`pixi run pip install -v -e '.[test,test_extras,msgpack,crc32c,pcodec,zfpy]'`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -238,3 +238,28 @@ warn_unreachable = false
 warn_redundant_casts = true
 warn_unused_ignores = true
 warn_unused_configs = true
+
+[tool.pixi.project]
+channels = ["conda-forge"]
+platforms = ["linux-64"]
+
+[tool.pixi.pypi-dependencies]
+numcodecs = { path = ".", editable = true }
+
+[tool.pixi.environments]
+default = { solve-group = "default" }
+crc32c = { features = ["crc32c"], solve-group = "default" }
+docs = { features = ["docs"], solve-group = "default" }
+msgpack = { features = ["msgpack"], solve-group = "default" }
+pcodec = { features = ["pcodec"], solve-group = "default" }
+test = { features = ["test"], solve-group = "default" }
+test-extras = { features = ["test_extras"], solve-group = "default" }
+zfpy = { features = ["zfpy"], solve-group = "default" }
+
+[tool.pixi.tasks]
+
+[tool.pixi.dependencies]
+clang = ">=19.1.7,<20"
+blosc = ">=1.21.6,<2"
+c-compiler = ">=1.9.0,<2"
+cxx-compiler = ">=1.9.0,<2"
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# GitHub syntax highlighting
+pixi.lock linguist-language=YAML linguist-generated=true
--- a/.gitignore
+++ b/.gitignore
@@ -104,3 +104,6 @@ numcodecs/version.py
 
 # Cython generated
 numcodecs/*.c
+# pixi environments
+.pixi
+*.egg-info
--- a/numcodecs/abc.py
+++ b/numcodecs/abc.py
@@ -29,14 +29,14 @@ other and vice versa.
 """
 
 from abc import ABC, abstractmethod
-from typing import Optional
+from typing import ClassVar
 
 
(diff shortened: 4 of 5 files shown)


d-v-b:
not sure if adding the blosc dep here is correct, since we are vendoring it (cc @jakirkham)

d-v-b:
cc @mkitti in case you have any advice

mkitti:
> not sure if adding the blosc dep here is correct, since we are vendoring it (cc @jakirkham)

I'll just point out that blosc is not even a dependency of the numcodecs package in conda-forge:
https://github.com/conda-forge/numcodecs-feedstock/blob/main/recipe/meta.yaml

I think this situation is a very weird. We're not even modifying the vendored blosc build here.

In contrast, the blosc package in conda-forge depends on lz4-c, snappy, zlib, and zstd packages, as this package should.
https://github.com/conda-forge/blosc-feedstock/blob/main/recipe/meta.yaml

That said, I have no idea how one would have those dependencies in PyPI via wheels.

martindurant:
cramjam, anyone?? This would just be the reason for them to slidify the blosc backend, but they have all the others in a statically-liked wheel

mkitti:
> statically-linked wheel

Is a big statically-linked wheel the ideal here?

martindurant:
> Is a big statically-linked wheel the ideal here?

If you're worried about dependencies, packaging and how to link to system libs, I suggest that yes. I note that cramjam is ~2MB. Also, zarr likes zarr crates, and cramjam can be called directly from rust, and also return python-consumable buffers.

mkitti:
I took a look at libcramjam. I can already see several major issues. For example, I do not see checksum options for Zstandard. Now in certain circumstances, perhaps this could be fine if we had a plugin system that allowed for multiple implementations of the same codec.

I think I would much rather have a direct dependency on libzstd rather than going through another intermediary.

pyproject.toml
```diff
@@ -238,3 +238,27 @@ warn_unreachable = false
...
+[tool.pixi.pypi-dependencies]
+numcodecs = { path = ".", editable = true }
+
+[tool.pixi.environments]
+default = { solve-group = "default" }
+crc32c = { features = ["crc32c"], solve-group = "default" }
+docs = { features = ["docs"], solve-group = "default" }
+msgpack = { features = ["msgpack"], solve-group = "default" }
+pcodec = { features = ["pcodec"], solve-group = "default" }
+test = { features = ["test"], solve-group = "default" }
```
mkitti, line 255:
```suggestion
test = { features = ["test", "crc32c", "msgpack", "pcodec", "zfpy"], solve-group = "default" }
```

Add optional dependencies for the test environment.

mkitti:
Rather than 

```
pixi run pip install -v -e '.[test,test_extras,msgpack,crc32c,pcodec,zfpy]'
```

I think either `pixi install` or `pixi reinstall` should be sufficient. Pixi should be able to handle all the dependencies. If the optional dependencies are desired, they could be inserted via features and (sub)environments.

`pixi reinstall` is relatively recent.

mkitti:
> I noticed the config has an editable install, just be aware there are two open Pixi issues about editable installs causing lockfile churn.

Why is this a problem? For a library like this, we could just add the lock file to .gitignore.

mkitti:
Add this the package as an editable install, remove the pip dependency, and do not recommend the use of `pixi run pip install -e `

pyproject.toml
```diff
@@ -244,13 +244,17 @@ warn_unused_configs = true
...
 
 [tool.pixi.dependencies]
 clang = ">=19.1.7,<20"
 c-compiler = ">=1.9.0,<2"
 cxx-compiler = ">=1.9.0,<2"
 hatch = '==1.14.1'
+pip = ">=25.2,<26"
+setuptools = ">=80.9.0,<81"
+cython = ">=3.1.4,<4"
+numpy = ">=2.3.3,<3"
```
mkitti, line 258:
```suggestion

[tool.pixi.pypi-dependencies]
numcodecs = { path = ".", editable = true }

```

mkitti (changes requested):

d-v-b:
this was handled by https://github.com/zarr-developers/numcodecs/pull/766

PR #709, opened 2025-02-14
Only skip failing test on macOS

dstansby:
I'm hoping this will allow the test to run and pass on Linux runners, and put our coverage up to 100% again so the PR list isn't full of ❌


dstansby:
pre-commit.ci autofix

PR #711, opened 2025-03-01
Make Checksum32 an ABC

dstansby:
Since it's not intended to be used without sub-classing, make `Checksum32` an ABC. This also means we can catch any sub-classes that don't implement `checksum()` more easily.

### DIFF
--- a/numcodecs/checksum32.py
+++ b/numcodecs/checksum32.py
@@ -1,11 +1,12 @@
+import abc
 import struct
 import zlib
-from collections.abc import Callable
 from contextlib import suppress
 from types import ModuleType
-from typing import TYPE_CHECKING, Literal, Optional
+from typing import Literal, Optional
 
 import numpy as np
+from typing_extensions import Buffer
 
 from .abc import Codec
 from .compat import ensure_contiguous_ndarray, ndarray_copy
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -13,11 +13,8 @@ name = "numcodecs"
 description = """
 A Python package providing buffer compression and transformation codecs \
 for use in data storage and communication applications."""
-readme =  "README.rst"
-dependencies = [
-    "numpy>=1.24",
-    "deprecated"
-]
+readme = "README.rst"
+dependencies = ["numpy>=1.24", "deprecated", "typing_extensions"]
```
jakirkham, line 17: Let's not unwrap this

jakirkham:
In the future @dstansby can you please request a review?

numcodecs/checksum32.py
```diff
@@ -92,9 +99,15 @@ class Adler32(Checksum32):
...
+    @staticmethod
+    def checksum(data: Buffer, value: int = 1) -> int:
+        """
+        Thin wrapper around ``zlib.adler32``.
+        """
+        return zlib.adler32(data, value)
+
 
 class JenkinsLookup3(Checksum32):
     """Bob Jenkin's lookup3 checksum with 32-bit output
```
jakirkham, line 113: It appears the `checksum` method was not added here

PR #712, opened 2025-03-01
Remove deprecated blosc code

dstansby:
This was deprecated in 0.15.0, as code that is either experimental or not intended to be public, so remove it for 0.16.0. This reduces the amount of Python interface we need to maintain for `blosc`, and enables the option to switch to using the third party blosc python package in the future.


d-v-b:
looks like this was a breaking change for some implementations. we might want to revert.

dstansby:
Sorry about this - there's a pin over at https://github.com/zarr-developers/zarr-python/pull/2965, which I'll try and get released ASAP

d-v-b:
I'm trying to understand why we didn't revert these changes before issuing a new release, as the changes in this PR are breaking for zarr-python 2.18.

dstansby:
It was my mistake that this broke zarr-python 2.18 - I think the easiest and most pragmatic fix is to put a pin in zarr-python 2.18, which I've proposed over at https://github.com/zarr-developers/zarr-python/pull/2965

d-v-b:
But given that this PR requires changes to zarr-python, why did we issue a [new numcodecs release](https://github.com/zarr-developers/numcodecs/releases/tag/v0.16.0) before the preparatory changes in zarr-python were in?

dstansby:
I definitely think we should discuss what process/development changes to make so this doesn't happen again, but in the short term it would be great to get a quick review and merge on https://github.com/zarr-developers/zarr-python/pull/2965 so we can unbreak zarr 2.18

d-v-b:
maybe i'm getting the timeline wrong, but it looks like the problems with this PR were reported yesterday, but the numcodecs release was today? In theory we could have deferred releasing these changes specifically until zarr-python was prepared for it.

dstansby:
Oh, I made the "release" entry no GitHub this morning, but the actual release/tagging was yesterday.

d-v-b:
gotcha, so my timeline was inaccurate

flying-sheep:
Over at scverse, we use some nice release automation to prevent anything from going wrong with coordinating GitHub/PyPI releases: https://github.com/scverse/anndata/blob/main/.github/workflows/publish.yml

Of course, if you hand-craft the release notes differently for the docs and GitHub, that’s not possible, but if you use the same text, it’s worth it I’d say.

PR #713, opened 2025-03-03
(feat): `typesize` declared with constructor for `Blosc`

ilan-gold:
See https://github.com/zarr-developers/zarr-python/issues/2766 and https://github.com/zarr-developers/zarr-python/issues/2171 - I checked locally the reproducer in the later for the compression ratio, and we now are getting the old performance with 1383 bytes

I think a good place to start is to declare the itemsize information, which is stored anyway on the `BloscCodec`

This PR does no validation of `typesize`, although it could in theory if the incoming buffer is a numpy array.  I'm not sure that's desireable though as this is purely a performance thing (i.e., a "wrong" `typesize` doesn't actually break compression/decompression). If you're so low-level that you're changing this while using arrays, and you get it wrong, you'd probably also notice that your files have bad compression ratios

TODO:

- [x] Tests pass locally
- [x] Docs build locally

### DIFF
--- a/numcodecs/blosc.pyx
+++ b/numcodecs/blosc.pyx
@@ -279,7 +279,9 @@ def compress(source, char* cname, int clevel, int shuffle=SHUFFLE,
     source_buffer = Buffer(source, PyBUF_ANY_CONTIGUOUS)
     source_ptr = source_buffer.ptr
     nbytes = source_buffer.nbytes
-    if typesize is not None:
+    if isinstance(typesize, int):
+        if typesize < 1:
+            raise ValueError(f"Cannot use typesize {typesize} less than 1.")
         itemsize = typesize
     else:
         itemsize = source_buffer.itemsize
@@ -572,6 +574,8 @@ class Blosc(Codec):
     max_buffer_size = 2**31 - 1
 
     def __init__(self, cname='lz4', clevel=5, shuffle=SHUFFLE, blocksize=AUTOBLOCKS, typesize=None):
+        if isinstance(typesize, int) and typesize < 1:
+            raise ValueError(f"Cannot use typesize {typesize} less than 1.")
         self.cname = cname
         if isinstance(cname, str):
             self._cname_bytes = cname.encode('ascii')
--- a/numcodecs/blosc.pyx
+++ b/numcodecs/blosc.pyx
@@ -280,7 +280,7 @@ def compress(source, char* cname, int clevel, int shuffle=SHUFFLE,
     source_ptr = source_buffer.ptr
     nbytes = source_buffer.nbytes
     if isinstance(typesize, int):
-        if typesize < 1:
+        if typesize < 1: # pragma: no cover
             raise ValueError(f"Cannot use typesize {typesize} less than 1.")
         itemsize = typesize
     else:
--- a/numcodecs/blosc.pyx
+++ b/numcodecs/blosc.pyx
@@ -280,7 +280,7 @@ def compress(source, char* cname, int clevel, int shuffle=SHUFFLE,
     source_ptr = source_buffer.ptr
     nbytes = source_buffer.nbytes
     if isinstance(typesize, int):
-        if typesize < 1: # pragma: no cover
+        if typesize < 1:
             raise ValueError(f"Cannot use typesize {typesize} less than 1.")
         itemsize = typesize
     else:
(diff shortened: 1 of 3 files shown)


d-v-b:
this a good idea, but we should be sure that we don't cause backwards compatibility problems with this, since it adds a key to zarr metadata. How do various zarr clients like zarr-python or tensorstore handle extra keys in the blosc JSON definition?

ilan-gold:
> this a good idea, but we should be sure that we don't cause backwards compatibility problems with this, since it adds a key to zarr metadata.

Why does it add a key? One of the reason I feel so secure doing this is precisely that I thought he were adding the key anyway: https://github.com/zarr-developers/zarr-python/blob/680142f6f370d862f305a9b37f3c0ce2ce802e76/src/zarr/codecs/blosc.py#L122-L136 so I felt extremely confident making this change.  Maybe I'm misunderstanding

ilan-gold:
This feels similar to the Zstd levels/checksum business - it's encode only, and can be ignored for reading back in. I understand why we serialize it, but it seems people sometimes ignore it anyway (thinking of the NVComp Zstd) with no effect (the utility/quality of such a decision can be debated probably)

numcodecs/blosc.pyx
```diff
@@ -235,7 +235,7 @@ def _err_bad_cname(cname):
 err_bad_cname = deprecated(_err_bad_cname)
 
 def compress(source, char* cname, int clevel, int shuffle=SHUFFLE,
-             int blocksize=AUTOBLOCKS):
+             int blocksize=AUTOBLOCKS, typesize=None):
```
dstansby, line 238: Is it worth mirroring the `blocksize` argument here, and instead of `None` defaulting to `0`, which will automatically set the type size if passed?

ilan-gold: `None` feels more pythonic than `0` - there is a c-blosc (i.e., the underlying C compressor) function `blosc_set_blocksize` that knows what to do with 0.  I'm don't think there is an equivalent for usage of this parameter (at least from my testing, giving `0` here actually writes out `0` to the underlying buffer).

ilan-gold: Adding something to disallow this

ilan-gold: > Perhaps worth adding a test here to see what happens when 0 or a negative value is passed?

I tried using `0` without the check I just added and got: 

```
>       codec_no_type_size.decode(encoded_without_itemsize)

numcodecs/tests/test_blosc.py:288: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
numcodecs/blosc.pyx:591: in numcodecs.blosc.Blosc.decode
    return decompress(buf, out)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   raise RuntimeError('error during blosc decompression: %d' % ret)
E   RuntimeError: error during blosc decompression: -1
```
i.e., `blosc` will write the value to disk no issue and then try to use it in decompression.  All the more important to disallow it up front (which I have added the check + test for)

dstansby (approved):

LDeakin:
> this a good idea, but we should be sure that we don't cause backwards compatibility problems with this, since it adds a key to zarr metadata. How do various zarr clients like zarr-python or tensorstore handle extra keys in the blosc JSON definition?

Indeed! This was a breaking change after all for implementations that reject unknown keys, as it seemingly was not serialised in the past. See this diff in https://github.com/LDeakin/zarrs/pull/174/files.

d-v-b:
ok so this _was_ actually a breaking change, that's my bad, I definitely should have reviewed this more carefully. But we also really need to be catching this with tests. 

@jbms @bogovicj sorry for the inconvenience but you might have to change how your implementations handle the blosc codec in zarr v2 data.

d-v-b:
it was released in [0.16](https://github.com/zarr-developers/numcodecs/releases/tag/v0.16.0)

PR #723, opened 2025-04-07
In `vlen`, define and use `const` `HEADER_LENGTH`

jakirkham:
Previously `HEADER_LENGTH` was defined in `vlen`, but not as `const`. So it could concievably be changed. While this didn't happen in practice, it does mean the compiler needs to generate code to check and use this value throughout.

However `HEADER_LENGTH` is intended to be `const`. So define it as such. That way the compiler can generate more efficient code when using it. Further define it in terms of the type whose size it references (namely `uint32_t`). Also make sure that type is `cimport`ed for referencing.

Finally consistently use `HEADER_LENGTH` when it applies throughout `vlen`. Previously `HEADER_LENGTH` was sometimes used and in other cases the value `4` was used. So consistently use `HEADER_LENGTH` throughout for improved code clarity. Now that `HEADER_LENGTH` is marked `const` this is equally efficient.

<hr>

TODO:

- [ ] Tests pass locally
- [ ] Docs build locally

### DIFF
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -14,6 +14,25 @@ Release notes
 
 .. _unreleased:
 
+Unreleased
+----------
+
+Enhancements
+~~~~~~~~~~~~
+
+Improvements
+~~~~~~~~~~~~
+
+* In ``vlen``, define and use ``const`` ``HEADER_LENGTH``.
+  By :user:`John Kirkham <jakirkham>`, :issue:`723`
+
+Fixes
+~~~~~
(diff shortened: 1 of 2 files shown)


d-v-b (approved):

jakirkham:
Weird would have thought automerge would have waited for the rest of the checks

Will keep an eye on `main` to confirm this does indeed pass (though it had earlier before resolving conflicts with the news entries)

Issue #735, opened 2025-04-09
Segfault in numcodecs/tests/test_vlen_bytes.py::test_encode_none


avalentino:
I run the unit test suite in debian sid (x86_64) i get a segmentation fault:
```
python3.13 -m pytes -v --pyargs numcodecs
============================= test session starts ==============================
platform linux -- Python 3.13.2, pytest-8.3.5, pluggy-1.5.0 -- /usr/bin/python3.13
cachedir: .pytest_cache
rootdir: /build/numcodecs-0.16.0+ds/.pybuild/cpython3_3.13_numcodecs/build
configfile: pyproject.toml
plugins: typeguard-4.4.2, cov-5.0.0
collecting ... collected 698 items / 27 deselected / 4 skipped / 671 selected

numcodecs/astype.py::numcodecs.astype.AsType PASSED                      [  0%]

[...]

numcodecs/tests/test_vlen_bytes.py::test_decode_errors PASSED            [ 96%]
numcodecs/tests/test_vlen_bytes.py::test_encode_none Fatal Python error: Aborted

Current thread 0x00007ebfefa2e100 (most recent call first):
  File "/build/numcodecs-0.16.0+ds/.pybuild/cpython3_3.13_numcodecs/build/numcodecs/tests/test_vlen_bytes.py", line 94 in test_encode_none
  File "/usr/lib/python3/dist-packages/_pytest/python.py", line 159 in pytest_pyfunc_call
  File "/usr/lib/python3/dist-packages/pluggy/_callers.py", line 103 in _multicall
  File "/usr/lib/python3/dist-packages/pluggy/_manager.py", line 120 in _hookexec
  File "/usr/lib/python3/dist-packages/pluggy/_hooks.py", line 513 in __call__
  File "/usr/lib/python3/dist-packages/_pytest/python.py", line 1627 in runtest
  File "/usr/lib/python3/dist-packages/_pytest/runner.py", line 174 in pytest_runtest_call
  File "/usr/lib/python3/dist-packages/pluggy/_callers.py", line 103 in _multicall
  File "/usr/lib/python3/dist-packages/pluggy/_manager.py", line 120 in _hookexec
  File "/usr/lib/python3/dist-packages/pluggy/_hooks.py", line 513 in __call__
  File "/usr/lib/python3/dist-packages/_pytest/runner.py", line 242 in <lambda>
  File "/usr/lib/python3/dist-packages/_pytest/runner.py", line 341 in from_call
[...]

Issue #683, opened 2025-01-07
macOS CI failing


dstansby:
macOS tests have started failing, with no obvious culprit. The last passing test run is [here](https://github.com/zarr-developers/numcodecs/actions/runs/12613206201?pr=571), and an example failing run [here](https://github.com/zarr-developers/numcodecs/actions/runs/12640624664/job/35221401153?pr=682).

The error is in `test_vlen_bytes.py::test_encode_none `:

```
 Fatal Python error: Aborted

Current thread 0x00007ff85e49a9c0 (most recent call first):
  File "/Users/runner/work/numcodecs/numcodecs/numcodecs/tests/test_vlen_bytes.py", line 90 in test_encode_none
  File "/Users/runner/miniconda3/envs/test/lib/python3.11/site-packages/_pytest/python.py", line 159 in pytest_pyfunc_call
  File "/Users/runner/miniconda3/envs/test/lib/python3.11/site-packages/pluggy/_callers.py", line 103 in _multicall
  File "/Users/runner/miniconda3/envs/test/lib/python3.11/site-packages/pluggy/_manager.py", line 120 in _hookexec
  File "/Users/runner/miniconda3/envs/test/lib/python3.11/site-packages/pluggy/_hooks.py", line 513 in __call__
  File "/Users/runner/miniconda3/envs/test/lib/python3.11/site-packages/_pytest/python.py", line 1627 in runtest
  File "/Users/runner/miniconda3/envs/test/lib/python3.11/site-packages/_pytest/runner.py", line 174 in pytest_runtest_call
  File "/Users/runner/miniconda3/envs/test/lib/python3.11/site-packages/pluggy/_callers.py", line 103 in _multicall
  File "/Users/runner/miniconda3/envs/test/lib/python3.11/site-packages/pluggy/_manager.py", line 120 in _hookexec
  File "/Users/runner/miniconda3/envs/test/lib/python3.11/site-packages/pluggy/_hooks.py", line 513 in __call__
  File "/Users/runner/miniconda3/envs/test/lib/python3.11/site-packages/_pytest/runner.py", line 242 in <lambda>
  File "/Users/runner/miniconda3/envs/test/lib/python3.11/site-packages/_pytest/runner.py", line 341 in from_call
  File "/Users/runner/miniconda3/envs/test/lib/python3.11/site-packages/_pytest/runner.py", line 241 in call_and_report
[...]

PR #736, opened 2025-04-09
Re-enable `VLenBytes` round-trip `None` test

jakirkham:
[Description of PR]

TODO:

- [ ] Tests pass locally
- [ ] Docs build locally


numcodecs/vlen.pyx
```diff
@@ -268,7 +271,7 @@ class VLenBytes(Codec):
             l = lengths[i]
             store_le32(<uint8_t*>data, l)
             data += HEADER_LENGTH
-            encv = PyBytes_AS_STRING(values[i])
```
jakirkham, line 271: Looks like this line was the actual one that failed ( a bit further in than the aforementioned error would suggest: https://github.com/zarr-developers/numcodecs/issues/683#issuecomment-2791172032 )

The issue is that `value[i]` does not contain the normalized value. So if `None` shows up here, it errors. Also the error is a nasty segfault (due to attempting to access a field that does not exist) or an `assert`ion error if Python's debug mode is used

numcodecs/vlen.pyx
```diff
@@ -240,6 +241,7 @@ class VLenBytes(Codec):
         n_items = values.shape[0]
 
         # setup intermediates
+        normed_values = np.empty(n_items, dtype=object)
```
jakirkham, line 244: To fix the error mentioned above, we need to create a place to store the normalized values

So we allocate an empty array to start. `object` type as we are storing Python objects in it

numcodecs/vlen.pyx
```diff
@@ -250,6 +252,7 @@ class VLenBytes(Codec):
                 b = b''
             elif not PyBytes_Check(b):
                 raise TypeError('expected byte string, found %r' % b)
+            normed_values[i] = b
```
jakirkham, line 255: Now when we loop through and normalize an input, we can store each normalized value for processing later

numcodecs/vlen.pyx
```diff
@@ -268,7 +271,7 @@ class VLenBytes(Codec):
             l = lengths[i]
             store_le32(<uint8_t*>data, l)
             data += HEADER_LENGTH
-            encv = PyBytes_AS_STRING(values[i])
+            encv = PyBytes_AS_STRING(normed_values[i])
```
jakirkham, line 274: Thus we can now access the normalized value here and know it is of the expected type

That said, this code can be improved further by introducing Cython typed variables and using them consistently to get better and more consistent type checking throughout

jakirkham (commented):
Have added some notes below to explain the cause of the bug and what was done here to fix it

jakirkham:
Antonio confirmed the bug fix: https://github.com/zarr-developers/numcodecs/issues/735#issuecomment-2791682750

Also have confirmed the fixed test on CI

Given that, will go ahead and put this fix in

PR #738, opened 2025-04-10
Unskip accidentally skipped tests

dstansby:
When a `pytest.skip()` is executed it skips the entire running test, not just the current for loop. This means putting a `pytest.skip()` inside a for loop can accidentally skip tests. The fix here is to parametrize the tests instead of using for loops.


jakirkham (approved):

jakirkham:
Thanks David! 🙏

PR #740, opened 2025-04-15
Use `pytest.skipif` to properly report skipped CRC32C codec tests

### LINKED ISSUE #640: Explicitly `pytest.skip` any tests that depend on optional dependencies
As a quick fix in https://github.com/zarr-developers/numcodecs/pull/637, some tests for the `CRC32C` codec are only ever defined if `crc32c` is installed. If `crc32c` isn't installed, these tests are silently not defined or reported to pytest.

In order to report that these tests are being skipped to pytest, it would be much better to mark them with a pytest skip mark, so when the dependency isn't installed these skipped tests get reported to pytest.

agriyakhetarpal:
The tests `test_err_location`, `test_repr`, and `test_backwards_compatibility` were not reporting that they were skipped if the `crc32c` dependency wasn't available/installed, as they weren't marked with the appropriate markers. This PR skips these tests at the `pytest` level rather than using a `has_crc32c` check that defines the tests at collection. I've parametrized most of these tests. I had to split up the `test_backwards_compatibility` test into two, as `CRC32C()` wouldn't be defined if it were not installed (I don't think it makes sense to mock it just for one test).

Closes #640

TODO:

- [x] Tests pass locally
- [ ] Docs build locally

### DIFF
--- a/numcodecs/tests/test_checksum32.py
+++ b/numcodecs/tests/test_checksum32.py
@@ -39,46 +39,51 @@ arrays = [
     np.random.randint(-(2**63), -(2**63) + 20, size=1000, dtype='i8').view('m8[m]'),
 ]
 
-codecs = [
+base_codecs = [
     CRC32(),
     CRC32(location="end"),
     Adler32(),
     Adler32(location="end"),
 ]
-if has_crc32c:
-    codecs.extend(
-        [
-            CRC32C(location="start"),
-            CRC32C(),
-        ]
-    )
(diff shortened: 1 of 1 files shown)


agriyakhetarpal:
🤔 I'm not sure if the Codecov report makes sense here. I've just changed `test_err_location` to use `if not has_crc32c` and skip it instead of using `if has_crc32c`. It should mean the same thing.

PR #741, opened 2025-04-22
add numcodecs.zarr3.to_zarr3 method

brokkoli71:
- implements `to_zarr3` function in `numcodecs.zarr3` for https://github.com/zarr-developers/zarr-python/issues/2964

TODO:

- [x] Tests pass locally
- [ ] Docs build locally

### DIFF
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -11,7 +11,6 @@ Release notes
     Unreleased
     ----------
 
-
 .. _unreleased:
 
 Unreleased
@@ -19,6 +18,9 @@ Unreleased
 
 Enhancements
 ~~~~~~~~~~~~
+
+* implement ``to_zarr3`` in ``numcodecs.zarr3`` to enable conversion of a codec to its zarr3-compatible equivalent.
+  By :user:`Hannes Spitz <brokkoli71>`
 
 Improvements
 ~~~~~~~~~~~~
@@ -63,7 +65,7 @@ Enhancements
 Removals
 ~~~~~~~~
 
-The following ``blosc`` funcitons are removed, with no replacement.
+The following ``blosc`` functions are removed, with no replacement.
 This is because they were not intended to be public API.
 
 - ``numcodecs.blosc.init``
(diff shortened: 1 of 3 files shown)


d-v-b:
Thanks for working on this! We should probably have a conversation about the strategy here. My preference would be to move _away_ from separate zarr 2 and zarr 3 codec classes, which would look somewhat different than the effort here.

rabernat:
> My preference would be to move _away_ from separate zarr 2 and zarr 3 codec classes, which would look somewhat different than the effort here.

Are these mutually exclusive?

I see this PR as a shim to solve a pretty urgent problem that Zarr users are experiencing in the V3 transition.

In the future, we could refactor how codec classes work, but that's likely a much slower process.

normanrz:
> > > My preference would be to move _away_ from separate zarr 2 and zarr 3 codec classes, which would look somewhat different than the effort here.
> > 
> > 
> > Are these mutually exclusive?
> > I see this PR as a shim to solve a pretty urgent problem that Zarr users are experiencing in the V3 transition.
> > In the future, we could refactor how codec classes work, but that's likely a much slower process.
> 
> One way to achieve this shim without adding more problematic zarr 2 / zarr 3 logic to numcodecs would be to implement the changes in this PR in zarr-python, instead of numcodecs. Is there any reason why that would not be possible?

I would argue that adding this to zarr-python actually increases the problematic coupling, because this `to_zarr3` method depends on private numcodecs interfaces. However, I think we can be pragmatic here and implement it on either side until we have resolved https://github.com/zarr-developers/numcodecs/issues/742

d-v-b:
> I would argue that adding this to zarr-python actually increases the problematic coupling, because this `to_zarr3` method depends on private numcodecs interfaces.

As numcodecs has so far existed chiefly for zarr-python's benefit, and we control numcodecs, I would argue that effectively all numcodecs interfaces are public to zarr-python. To put it differently, "zarr-python uses numcodecs interface X" would be a valid reason for us not to change that interface, whether interface X was public or not. 

This is of course a problematic, and ultimately something we should fix. I think the first steps would be to fully extract as much zarr-specific-logic from numcodecs, which argues for making the code in this PR over in zarr-python.

PR #766, opened 2025-07-17
Handle zarr 3.1.0

d-v-b:
Implements a data type adapter layer for `numcodecs.zarr3`. Ensures that zarr data structures get the right data type.

this hopefully fixes the breakage caused by the zarr 3.1 release.

TODO:

- [ ] Tests pass locally
- [ ] Docs build locally

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -91,7 +91,7 @@ jobs:
         uses: prefix-dev/setup-pixi@v0.8.14
         with:
           pixi-version: v0.49.0
-          cache: true
+          cache: false
 
       - name: List deps
         shell: "bash -l {0}"


dstansby:
For now, can we not just have one test run on the GH Actions CI that tests against `zarr-python` 3.0.x?

d-v-b:
> For now, can we not just have one test run on the GH Actions CI that tests against `zarr-python` 3.0.x?

sure! can you write it?

d-v-b:
I added pixi + hatch to pyproject.toml exclusively for running the zarr-python tests against two different versions of zarr-python. We can always add to this later.

pixi.lock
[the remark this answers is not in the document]
dstansby: What's the time difference in this case when creating a new environment with vs. without a lockfile?

> Trying to unpick why a build failed without a known good-state can be a nightmare.

Agreed - I've found that printing out the environment as part of the test run works for this.

I'd prefer not to commit a lockfile, but if we do, we should make sure at the same time there is an automated method set up to update it so it doesn't get out of date without us really realising.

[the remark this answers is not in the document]
d-v-b: If we don't commit the lockfile, then should we add it to gitignore?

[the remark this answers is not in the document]
d-v-b: i removed the lockfile and added `pixi.lock` to gitignore

numcodecs/zarr3.py
```diff
@@ -56,6 +58,23 @@
 CODEC_PREFIX = "numcodecs."
 
 
+def _from_zarr_dtype(dtype: Any) -> np.dtype:
```
TomNicholas, line 61: Is the input type here really `Any`?

numcodecs/zarr3.py
```diff
@@ -56,6 +58,23 @@
...
+    """
+    Get a numpy data type from an array spec, depending on the zarr version.
+    """
+    if Version(version('zarr')) >= Version("3.1.0"):
+        return dtype.to_native_dtype()
+    return dtype  # pragma: no cover
+
+
+def _to_zarr_dtype(dtype: np.dtype) -> Any:
```
TomNicholas, line 70: Same here - shouldn't it only return subclasses of zarr data types?

TomNicholas: https://github.com/zarr-developers/zarr-python/blob/a26926cf4ec51a12ca28ff31c13ef2c8a32d51f9/src/zarr/core/dtype/__init__.py#L191

TomNicholas (approved):
This is important to get in because currently there is zarr data written in the wild that cannot be read with recent zarr-python versions without this PR.

numcodecs/zarr3.py
d-v-b: no, but see https://github.com/zarr-developers/numcodecs/pull/766#discussion_r2270237827, it's not easy to annotate this type properly. And IMO we don't really need type safety for these functions right now, because they are very narrowly used.

TomNicholas: Fair enough.

PR #777, opened 2025-08-12
Fix to work with Zarr 3.1.0

dstansby:
This is an attempt at a minimal version of https://github.com/zarr-developers/numcodecs/pull/766, without the extra pixi stuff, which is orthogonal to fixing the bug.


dstansby:
@d-v-b this is a 'minimised' version of your PR, which uses the existing test infastructure instead of introducing `pixi`, which can be done in an orthogonal PR. Perhaps you could a review, and then we can get `numcodecs` development up and running again.

PR #780, opened 2025-08-13
chore/remove zarr3

d-v-b:
this PR removes the classes defined in `numcodecs.zarr3`, and instead re-exports those classes from `zarr.codecs._numcodecs`. This PR is a draft until https://github.com/zarr-developers/zarr-python/pull/3376 is merged in zarr-python.

### DIFF
--- a/numcodecs/zarr3.py
+++ b/numcodecs/zarr3.py
@@ -25,330 +25,44 @@ You can use codecs from :py:mod:`numcodecs` by constructing codecs from :py:mod:
 
 from __future__ import annotations
 
-import asyncio
-import math
-from dataclasses import dataclass, replace
-from functools import cached_property
 from importlib.metadata import version
-from typing import Any, Self
-from warnings import warn
 
-import numpy as np
 from packaging.version import Version
-
-import numcodecs
 
 try:
     import zarr  # noqa: F401
 
-    if Version(version('zarr')) < Version("3.0.0"):  # pragma: no cover
-        raise ImportError("zarr 3.0.0 or later is required to use the numcodecs zarr integration.")
+    zarr_version = version('zarr')
+    if Version(zarr_version) < Version("3.0.8"):  # pragma: no cover
+        msg = f"zarr 3.0.9 or later is required to use the numcodecs zarr integration. Got {zarr_version}."
+        raise ImportError(msg)
 except ImportError as e:  # pragma: no cover
-    raise ImportError(
-        "zarr 3.0.0 or later is required to use the numcodecs zarr integration."
-    ) from e
-
-from zarr.abc.codec import ArrayArrayCodec, ArrayBytesCodec, BytesBytesCodec
-from zarr.abc.metadata import Metadata
-from zarr.core.array_spec import ArraySpec
-from zarr.core.buffer import Buffer, BufferPrototype, NDBuffer
-from zarr.core.buffer.cpu import as_numpy_array_wrapper
-from zarr.core.common import JSON, parse_named_configuration, product
-
-CODEC_PREFIX = "numcodecs."
-
-
-def _from_zarr_dtype(dtype: Any) -> np.dtype:
-    """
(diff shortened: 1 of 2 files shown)


d-v-b:
A summary of this PR:
- removes zarr-python-dependent codec classes, since they are now defined in `zarr.codecs.numcodecs`. This change is not breaking because we still export these codecs from `numcodecs.zarr3`, but importing from `numcodecs.zarr3` will generate a deprecation warning:
 ```python
>>> from numcodecs.zarr3 import Blosc
<stdin>-1:1: DeprecationWarning: The numcodecs.zarr3 module is deprecated and will be removed in a future release of numcodecs. Import Blosc via zarr.codecs.numcodecs.Blosc instead. This requires Zarr Python >= 3.1.3. 
```
- removes tests for the codecs that are now defined in `zarr.codecs.numcodecs`, because these tests have been ported to Zarr Python/
- removes zarr installation from our main CI workflow
- removes hatch from pyproject.toml, uses uv instead. This was necessary to get around an obscure, hard-to-debug test failure. Instead of using hatch's dependency matrix support, we declare two separate, conflicting dependency groups and use UV to run tests in each one. The actual matrix is defined in a github action.
- CI tests against zarr python 3.1.2 (to ensure that importing from `numcodecs.zarr3` fails completely) and zarr python 3.1.3 (to ensure that import from `numcodecs.zarr3` succeeds, but with a deprecation warning)

**Comment by maxrjones** [comment]
When I add this PR as a dependency in VirtualiZarr while using an older version of Zarr (v3.1.0), I get this error:

```bash
ImportError while loading conftest '/Users/max/Documents/Code/zarr-developers/VirtualiZarr/conftest.py'.
conftest.py:17: in <module>
    from virtualizarr.manifests import ChunkManifest, ManifestArray
virtualizarr/__init__.py:3: in <module>
    from virtualizarr.accessor import (
virtualizarr/accessor.py:17: in <module>
    from virtualizarr.manifests import ManifestArray
virtualizarr/manifests/__init__.py:4: in <module>
    from virtualizarr.manifests.array import ManifestArray  # type: ignore # noqa
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
virtualizarr/manifests/array.py:6: in <module>
    from zarr.core.metadata.v3 import ArrayV3Metadata, RegularChunkGrid
.pixi/envs/upstream/lib/python3.13/site-packages/zarr/__init__.py:2: in <module>
    from zarr.api.synchronous import (
.pixi/envs/upstream/lib/python3.13/site-packages/zarr/api/synchronous.py:7: in <module>
    import zarr.api.asynchronous as async_api
.pixi/envs/upstream/lib/python3.13/site-packages/zarr/api/asynchronous.py:13: in <module>
    from zarr.core.array import (
.pixi/envs/upstream/lib/python3.13/site-packages/zarr/core/array.py:30: in <module>
    from zarr.codecs._v2 import V2Codec
.pixi/envs/upstream/lib/python3.13/site-packages/zarr/codecs/__init__.py:5: in <module>
    from zarr.codecs.crc32c_ import Crc32cCodec
.pixi/envs/upstream/lib/python3.13/site-packages/zarr/codecs/crc32c_.py:8: in <module>
    from crc32c import crc32c
E   ModuleNotFoundError: No module named 'crc32c'
```

maxrjones:
> > Should Zarr now be an optional dependency for numcodecs? It's currently only listed in dependency-groups as far as I see, which means it's not included as an optional dependency on release distributions

> My feeling is that Zarr should not be listed as an optional dependency, because numcodecs doesn't actually use Zarr for anything. But as we are in a bit of a weird situation, maybe there is some reason to declare Zarr as an optional dep here? Open to ideas.
> Wouldn't an optional dep be the circular dependency we want to get rid of?
> 

makes sense, thanks for explaining

jhamman (approved):

PR #784, opened 2025-08-19
Test 32-bit support via x86 Alpine Setup Action Linux

mkitti:
Test compilation in 32-bit Linux via Setup Alpine Linux:
https://github.com/marketplace/actions/setup-alpine-linux-environment

Currently does not support zfp or pcodec


d-v-b:
@mkitti is there any reason why we shouldn't merge this?

mkitti:
This looks good to merge to me.

I can iteration on it later to expand optional dependencies and maybe some small reorganization.

PR #835, opened 2026-04-12
chore: update pre-commit hook mypy

DimitriPapadopoulos:
updates:
- github.com/pre-commit/mirrors-mypy: v1.16.0 → v2.1.0

TODO:

- [x] Tests pass locally
- [ ] Docs build locally

### DIFF
--- a/tests/test_zarr3.py
+++ b/tests/test_zarr3.py
@@ -5,9 +5,9 @@ from typing import TYPE_CHECKING
 import pytest
 
 if TYPE_CHECKING:  # pragma: no cover
-    import zarr
+    import zarr.codecs
 else:
-    zarr = pytest.importorskip("zarr", "3.1.3")
+    zarr = pytest.importorskip("zarr.codecs", "3.1.3")
 
 import numcodecs.zarr3 as zarr3
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -25,9 +25,9 @@ repos:
     hooks:
     - id: sp-repo-review
 
-  -   repo: https://github.com/pre-commit/mirrors-mypy
-      rev: v1.16.0
-      hooks:
-      -   id: mypy
-          args: [--config-file, pyproject.toml]
-          additional_dependencies: [numpy, pytest, google-crc32c, crc32c, zfpy, 'zarr>=3.1.3']
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v1.20.0
+    hooks:
+    - id: mypy
+      args: [--config-file, pyproject.toml]
+      additional_dependencies: [numpy, pytest, google-crc32c, crc32c, zfpy, 'zarr>=3.1.3']
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -25,9 +25,9 @@ repos:
     hooks:
     - id: sp-repo-review
 
-  -   repo: https://github.com/pre-commit/mirrors-mypy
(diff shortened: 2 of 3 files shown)


tests/test_zarr3.py
```diff
@@ -45,4 +45,4 @@ def test_export(codec_name: str) -> None:
         DeprecationWarning,
         match="The numcodecs.zarr3 module is deprecated and will be removed in a future release of numcodecs. ",
     ):
-        assert getattr(zarr3, codec_name) == getattr(zarr.codecs.numcodecs, codec_name)
+        assert getattr(zarr3, codec_name) == getattr(zarr.codecs.numcodecs, codec_name)  # type: ignore[attr-defined]
```
d-v-b: why do we need a new type: ignore here? Is this improved type safety (if so, how?) or is this a regression in mypy

DimitriPapadopoulos: I think this is a regression, but I often have a hard time understanding Python typing. The error can be seen in https://results.pre-commit.ci/run/github/68207196/1775503887.A9wQYmi0SzyJZg-CuQfz6A:
```
tests/test_zarr3.py:48: error: Module has no attribute "codecs"
```

I understand MyPy believes that the `zarr>=3.1.3` module does not have a `codecs` submodule. I don't see why:
https://github.com/zarr-developers/zarr-python/tree/de947648e6ab6e7079ddb4f33b6b56070c3e6657/src/zarr/codecs

d-v-b: if the only effect of this version bump is a new mypy bug to mitigate, then maybe we wait for 1.20.1 😅

DimitriPapadopoulos: Let me find which exact version introduces this issue.

DimitriPapadopoulos: Strange, I cannot reproduce this issue outside pre-commit, instead I get other similar and nonsensical error messages with MyPy ≥ 1.19.0:
```console
$ uvx mypy@v1.18.2 tests
Success: no issues found in 39 source files
$ 
$ uvx mypy@v1.19.0 tests
tests/test_checksum32.py:19: error: Module has no attribute "util"  [attr-defined]
tests/test_checksum32.py:20: error: Module has no attribute "util"  [attr-defined]
Found 2 errors in 1 file (checked 39 source files)
$ 
$ uvx mypy@v1.20.0 tests
tests/test_checksum32.py:19: error: Module has no attribute "util"  [attr-defined]
tests/test_checksum32.py:20: error: Module has no attribute "util"  [attr-defined]
Found 2 errors in 1 file (checked 39 source files)
$ 
```
The module is `importlib`:
https://github.com/zarr-developers/numcodecs/blob/1fdde70aaaa1429cb84a5e14d2637fa498b24a28/tests/test_checksum32.py#L19-L20
and of course [`importlib`](https://docs.python.org/3/library/importlib.html) does have the [`util`](https://docs.python.org/3/library/importlib.html#module-importlib.util) submodule.

I fail to  understand why MyPy emits different errors within pre-commit and outside pre-commit.

DimitriPapadopoulos: Meanwhile, let's bump to 1.18.2.

DimitriPapadopoulos: It's easy to fix for MyPy ≥ 1.19.0:
```diff
-    import zarr.
+    import zarr.codecs
```
Not sure it makes sense, but it doesn't hurt.

I still fail to understand why MyPy emits different errors within pre-commit and outside pre-commit.

Issue #810, opened 2025-11-24
`RuntimeError: error during blosc decompression: -1`


alxmrs:
When running a large data processing job involving regridding oceans datasets on Dask distributed (via coiled), I keep hitting this error intermittently: 

```
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/Users/alxmrs/git/ocean_emulators/ocean_emulators/__main__.py", line 469, in <module>
    self._collect(ds)
    ~~~~~~~~~~~~~^^^^
  File "/Users/alxmrs/mamba/envs/ocean_emulators/lib/python3.13/site-packages/fire/core.py", line 135, in Fire
    component_trace = _Fire(component, args, parsed_flag_args, context, name)
  File "/Users/alxmrs/mamba/envs/ocean_emulators/lib/python3.13/site-packages/fire/core.py", line 468, in _Fire
    component, remaining_args = _CallAndUpdateTrace(
                                ~~~~~~~~~~~~~~~~~~~^
        component,
        ^^^^^^^^^^
    ...<2 lines>...
        treatment='class' if is_class else 'routine',
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        target=component.__name__)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/alxmrs/mamba/envs/ocean_emulators/lib/python3.13/site-packages/fire/core.py", line 684, in _CallAndUpdateTrace
    component = fn(*varargs, **kwargs)
  File "/Users/alxmrs/git/ocean_emulators/ocean_emulators/__main__.py", line 460, in om4
    dims_to_keep.append("ilev")
    ^^^^^^^^^
  File "/Users/alxmrs/git/ocean_emulators/ocean_emulators/__main__.py", line 245, in _collect
    ds.to_zarr(
    ~~~~~~~~~~^
        self.output_path,
        ^^^^^^^^^^^^^^^^^
    ...<2 lines>...
        compute=True,
        ^^^^^^^^^^^^^
    )
    ^
  File "/Users/alxmrs/mamba/envs/ocean_emulators/lib/python3.13/site-packages/xarray/core/dataset.py", line 2377, in to_zarr
    return to_zarr(  # type: ignore[call-overload,misc]
        self,
    ...<16 lines>...
[...]

PR #843, opened 2026-06-05
Reproduce #810

alxmrs:
Claude was able to diagnose and properly create a reproduction of #810. This is it's hypothesis and a script to explain what the root cause is. The implication is that blosc is doing the correct thing here, but there might be something else we could do in the Zarr-Python ecosystem to make Zarr robust to this class of error.


d-v-b:
the basic issue seems to be that the blosc codec is given invalid data? this seems like something the transport layer should be checking. I don't think there are any actionable codec changes to address that, because "invalid input data" is kind of an open space of problems.

alxmrs:
I agree with your takeaway @d-v-b. Happy to close the original issue as a "won't fix". I think the workaround listed there is sufficient to help future people deal with this error.

PR #844, opened 2026-06-18
chore: pin pytest and dedupe dependency groups

d-v-b:
pins pytest to an exact version (which resolves recent test failures) and deduplicates our dependency group specification.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -87,22 +87,20 @@ dev = [
     "cython>=3.0",
     "setuptools-scm>=6.2",
     "numpy>=2",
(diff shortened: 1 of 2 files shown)


d-v-b:
i'm leaving the pytest pin, but this PR also requires changes to our test fixtures to avoid deprecation warnings in pytest 9.x. those changes do not reduce test efficacy.

d-v-b:
the i386 test will be failing, but that's resolved by #845, which I will merge next. So I'm going to merge this despite checks failing.

PR #857, opened 2026-09-07
ci: Add Windows ARM64 build

ndabas:
As the title says. A couple of points to note:
- On Windows builds, meson was picking up the MSYS2 MINGW64 toolchain in preference to the MSVC one. That works on x86_64 but fails on arm64 because that toolchain targets x86_64 (there is no mingw equivalent for arm64). I have added the ARM64 build in a way that only that build uses MSVC, the Windows x86_64 build is as-is for now.
  - However as the MINGW64 environment is deprecated anyway, it might be best to either switch the x86_64 build to MSVC as well, or use the CLANG64/CLANGARM64 environments for consistency.
 - The zfpy package does not publish wheels for win_arm64, and those wheels don't build from source either as the sdist doesn't build `zfp.lib`. So I have skipped installing that for tests on Windows ARM64.

Let me know if you would like me to change anything on these points.

The CI passes on my fork, although it looks like a disaster because all of the codecov steps and the jobs that need large macos runners fail.


mkitti:
I thought MSYS2 has preliminary arm64 support:
https://www.msys2.org/docs/arm64/

I would prefer clang64 and clangarm64 if possible. MSVC is going to introduce new headaches.

ndabas:
@mkitti thanks for the review -- I accidentally closed this PR when playing around with branches and CI -- I have made the changes to use MSYS2 clang in #861, if you could review that. Thanks!

mkitti:
Did you want to reopen it?

ndabas:
Whatever your preference is -- I already opened a new one in #861 -- we can continue the review there if you wish.

Issue #749, opened 2025-05-26
ENH: Free-Threading (No-GIL) Support for Python 3.13+


HaoZeke:
xref https://github.com/zarr-developers/numcodecs/issues/670

I took a quick look, and I'm not wholly sure on what parts of the locking, possibly synchronizing access to the c-bloc library. Worst case scenario, maybe each of the codecs need to be checked for support.

The standard TODOs for adding free-threading support are in two phases, the first of which is a baseline consisting of the following tasks:

- [ ] Run the test suite with `pytest-run-parallel` to find potential issues, and fix them.
- [ ] Run the test suite under ThreadSanitizer. _If possible, depends on how many dependencies there are and if they run under TSan._
- [ ] Add `cp313t-*` to CI to build free-threading wheels.

The next phase is a bit more invasive, and it involves a single task:

- [ ] Audit Python bindings and declare them free-threading compatible (xref https://py-free-threading.github.io/porting/#updating-extension-modules).

For more details, please see the [suggested plan of attack in the py-free-threading guide](https://py-free-threading.github.io/porting/#suggested-plan-of-attack). The two phase approach is discussed in https://github.com/explosion/cymem/issues/47#issuecomment-2894478856 and https://github.com/gorakhargosh/watchdog/issues/1105#issuecomment-2888439578.

Note that this is the first time I've looked at this repo, so I might be
missing known issues or code that needs closer inspection. Any suggestions here
will be very useful. I'll start with the build without warning PR as also discussed in #670.

Let me know if I ought to hold off for any reason. (including if @tomwhite is working on it).

HaoZeke:
One of the issues is `zfpy`

```
DEBUG No compatible version found for: numcodecs[zfpy]
  × No solution found when resolving dependencies:
  ╰─▶ Because only the following versions of zfpy are available:
          zfpy<=1.0.0
          zfpy==1.0.0.post1
          zfpy==1.0.0.post3
          zfpy==1.0.1
      and zfpy==1.0.0 has no wheels with a matching Python implementation tag (e.g., `cp313`), we can conclude that zfpy>=1.0.0,<1.0.0.post1 cannot be used.
      And because zfpy>=1.0.0.post1,<=1.0.0.post3 was yanked (reason: arm64 wheels still broken), we can conclude that zfpy>=1.0.0.post1,<=1.0.0.post3 cannot
      be used.
[...]

PR #858, opened 2026-09-09
Add free-threaded CPython (3.14t) support

kumaraditya303:
Hello, I am a CPython Core Developer working on free-threading support. This PR adds support for free-threading on 3.14t+ (https://github.com/zarr-developers/numcodecs/issues/749). 

Changes:
- Mark all Cython extensions with freethreading_compatible=True so the interpreter no longer re-enables the GIL when numcodecs is imported
- Require Cython>=3.1, needed for the freethreading_compatible directive
- blosc: replace the lazily created multiprocessing.Lock (and get_mutex()) with a module-level threading.Lock, reset in forked children via os.register_at_fork
- blosc: take the lock in _init, _destroy, set_nthreads and the global context path of decompress, so blosc_set_nthreads cannot re-create the global context while another thread is using it
- blosc: compute the compressor list once at import; blosc_list_compressors() fills a static buffer without locking and was called on every compress()
- CI: add 3.14t to the test matrix, installing without pcodec, zfpy and google_crc32c which do not ship free-threaded wheels yet
- Wheels: enable cpython-freethreading in cibuildwheel, build cp314t-* wheels, and assert in the wheel test that the GIL is not re-enabled on free-threaded builds

TODO:

- [x] Tests pass locally
- [ ] Docs build locally

### DIFF
--- a/src/numcodecs/blosc.pyx
+++ b/src/numcodecs/blosc.pyx
@@ -112,7 +112,8 @@ def list_compressors():
 def get_nthreads():
     """Get the number of threads that Blosc uses internally for compression and
     decompression."""
-    return blosc_get_nthreads()
+    with _MUTEX:
+        return blosc_get_nthreads()
 
 
 def set_nthreads(int nthreads):
--- a/src/numcodecs/blosc.pyx
+++ b/src/numcodecs/blosc.pyx
@@ -3,6 +3,7 @@
 # cython: linetrace=False
 # cython: binding=False
 # cython: language_level=3
+# cython: freethreading_compatible=True
 import threading
 import multiprocessing
 import os
@@ -72,22 +73,20 @@ AUTOSHUFFLE = -1
 # automatic block size - let blosc decide
 AUTOBLOCKS = 0
 
-# synchronization
-_MUTEX = None
-_MUTEX_IS_INIT = False
+# computed once at import; blosc_list_compressors() fills a static buffer without locking
+_COMPRESSORS = tuple(blosc_list_compressors().decode('ascii').split(','))
 
-def get_mutex():
-    global _MUTEX_IS_INIT, _MUTEX
-    if not _MUTEX_IS_INIT:
-        try:
-            mutex = multiprocessing.Lock()
-        except OSError:
-            mutex = None
-        except ImportError:
-            mutex = None
-        _MUTEX = mutex
-        _MUTEX_IS_INIT = True
-    return _MUTEX
+# serializes access to blosc's global context
+_MUTEX = threading.Lock()
+
+
+def _reset_mutex_after_fork():
+    global _MUTEX
+    _MUTEX = threading.Lock()
+
+
+if hasattr(os, 'register_at_fork'):
+    os.register_at_fork(after_in_child=_reset_mutex_after_fork)
 
 # store ID of process that first loads the module, so we can detect a fork later
 _importer_pid = os.getpid()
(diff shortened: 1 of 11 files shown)


kumaraditya303:
I dropped `-l` from shell so that setup-python's PATH order is respected and free-threaded Python is installed correctly, otherwise it incorrectly picks up gil enabled python on macos 3.14t.

src/numcodecs/blosc.pyx
ngoldbaum: doesn't this need locking too?

d-v-b:
I don't know this part of the code very well so I'm acting as a conduit for Claude here. This is the review from claude:

Built against free-threaded 3.14.7 locally (macOS arm64, Cython 3.3):

- `import numcodecs` leaves the GIL disabled (`sys._is_gil_enabled()` is `False`).
- Full test suite passes on 3.14t (721 passed).
- Stress loop: 8 threads doing `compress`/`decompress` on the global context (`use_threads = True`) while a 9th thread loops `set_nthreads(1..4)`, run with `MallocScribble=1`. This branch is clean; `main` segfaults within a few seconds. Cause: `blosc_set_nthreads` frees and re-allocates `g_global_context` and `global_comp_mutex` without holding the mutex, so c-blosc's own locking cannot protect a concurrent `blosc_compress`. Taking the Python lock in `_init`/`_destroy`/`set_nthreads` and around the two global-context call sites is the right shape. The `*_ctx` paths read no mutable globals, so leaving them unlocked is correct. Caching `blosc_list_compressors()` at import is correct given the static buffer and unguarded "done" flag.

Required before merge:

1. **Release note** in `docs/release.md`. Besides free-threading support, two user-visible changes need spelling out: `numcodecs.blosc.get_mutex` is removed, and on platforms where `multiprocessing.Lock()` used to raise (no `sem_open`), the single-threaded ctx path was previously forced and blosc's internal threads will now be used from the main thread.

2. **The 3.14t CI jobs do not verify the GIL stays disabled.** Only the cibuildwheel test command does. If a later change re-enabled the GIL, `pytest` on 3.14t would still be green. A test that runs `import numcodecs` in a subprocess and asserts `not sys._is_gil_enabled()` when `sysconfig.get_config_var("Py_GIL_DISABLED")` is set would close that gap.

3. **Regression test for the locking.** The stress loop above fails on `main` and takes ~1s. Sketch:

   ```python
   def test_set_nthreads_races_global_context():
       data = np.arange(200_000, dtype='i4')
       stop = threading.Event()
       errors = []

       def worker():
           try:
               while not stop.is_set():
                   enc = blosc.compress(data, b'lz4', 5)
                   assert blosc.decompress(enc) == data.tobytes()
           except Exception as e:
               errors.append(e)

       def toggler():
           n = 1
           while not stop.is_set():
               n = 1 + (n % 4)
               blosc.set_nthreads(n)

       blosc.use_threads = True
       try:
           threads = [threading.Thread(target=worker) for _ in range(8)]
           threads.append(threading.Thread(target=toggler))
           for t in threads:
               t.start()
           time.sleep(1)
           stop.set()
           for t in threads:
               t.join()
       finally:
           blosc.use_threads = None
       assert not errors

d-v-b:
"I" can make the required changes if you don't mind

kumaraditya303:
> "I" can make the required changes if you don't mind

Yes you can make the changes, they all seem fine to me.

d-v-b (approved):

PR #860, opened 2026-09-18
Fix dependencies for blosc extension

QuLogic:
The blosc extension only depends on `blosc.h`, not any of the other libraries' headers. And if we're using internal blosc, then the other libraries are handled by `blosc_lib`.

Additionally, Zlib is only required if building blosc internally, but also add an option to allow using a system copy of it if desired.

- [x] Tests pass locally
- [n/a] Docstrings and API docs for any new/modified user-facing classes and functions
- [n/a] Docs build locally


d-v-b:
thanks, this looks good. I requested review from @mkitti who knows this stuff better than me. if that takes too long I will ask an llm.

mkitti:
This also looks good to me. I think this is scoped to meson though. Is that the only build method we support?

I would just add a note that this only affects thr meson build path.

mkitti (approved):
Nevermind. The build system was switched to meson.

PR #861, opened 2026-09-18
ci: Add Windows ARM64 build

ndabas:
As the title says. [Take 2](https://github.com/zarr-developers/numcodecs/pull/857).

- Switched from MINGW64 to CLANG64/CLANGARM64, as I suggested in my original PR, also mentioned by @mkitti in https://github.com/zarr-developers/numcodecs/issues/857#issuecomment-5727995509.
  - However because of [a bug in the Python headers](https://github.com/python/cpython/issues/157212), we cannot build for 3.14t on win_arm64 currently. win_amd64 works fine though.
- The zfpy package does not publish wheels for win_arm64, and those wheels don't build from source either as the sdist doesn't build zfp.lib. So I have skipped installing that for tests on Windows ARM64.


mkitti:
Is the plan to proceed without 3.14t wheels then?

ndabas:
Well there is a bug (python/cpython/issues/157212) in the Python 3.14t headers that specifically affects clang on MSYS2 CLANGARM64. If they fix that issue we will be able to build Windows ARM64 wheels for 3.14t.

The other option is to switch to MSVC which works fine in this case.

Your call really -- I have both code paths ready.

